PR #6, opened 2020-04-09
Feature/subproject setup

Baughn:
Set up a root build file collecting the subproject(s), remove the broken mc integration code, and add some basic tests.


Baughn:
This freezes the current output of the tests. Probably they'll change! If so, then there should in fact be an application available to rewrite the testdata, but I'm not doing that now.

Baughn:
Also, basicCircuitTest doesn't in fact contain any assertions, so that's... not a test. I'll have to ask you to fix that.

.gitignore
```diff
@@ -32,9 +33,3 @@ options.txt
 *.csproj.user
 *.sln
 App.config
-
-# Build artifacts
-*/obj/
-
-# Gnuplot files
-*.dat
```
jrddunbr, line 40: I added this and we should add it back

jrddunbr:
Oh, derp. Graham informs me that's actually used for output comparison tests.

PR #9, opened 2020-04-09
Adds Gnuplot data file parser

jrddunbr:
This could probably be used to make our test cases more sane?

Honestly, I used doubly linked lists because it was easy, but it may be worth considering converting it to some kind of array. In the least, it should be very fast since I used an iterator on the lists and not array indices.

I should also write tests for this to make sure that it actually works the way it should. I wrote this without running it once and I presume it works...


jrddunbr:
LGTM

PR #17, opened 2020-04-10
Change the testing infrastructure

Grissess:
Previously, tests were mostly just data-generators; if we're going to switch to a nice unit-testing framework, we should break the tests into units.

### DIFF
--- a/core/src/main/kotlin/org/eln2/sim/electrical/mna/Circuit.kt
+++ b/core/src/main/kotlin/org/eln2/sim/electrical/mna/Circuit.kt
@@ -1,6 +1,7 @@
 package org.eln2.sim.electrical.mna
 
 import org.apache.commons.math3.linear.*
+import org.eln2.debug.dprint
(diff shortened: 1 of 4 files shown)


jrddunbr:
I think that continuing to work on the gnuplot testing framework will also be important, because I really want to have confidence that the amount of energy that goes into the capacitors can come out of them, and have some relatively realistic curves. That may require some interesting math to do checks for.

jrddunbr (approved):
LGTM

jrddunbr:
We may want to not try to circumvent the fail for the one task that is a known bug, but I'm not sure on that. I think @Baughn has some thoughts on testing best practice that may help here.

Grissess:
Worth noting: the capacitors and inductors generally require identical timesteps for the integration to work correctly. I seem to recall that I did not use the (possibly slightly more accurate in some circumstances) triangular integration because it had worse stability.

The test which is presently failing is worked around at the moment.

PR #22, opened 2020-04-12
Add Codacy Coverage

jrddunbr:
This is an attempt to add the Codacy Coverage tool. It could backfire spectacularly.


jrddunbr:
I have set the GitHub Actions secret string so it *should* just pull that and work, no guarantees.

PR #29, opened 2020-04-13
Adds UMC to the Minecraft mod environment.

### LINKED ISSUE #16: Minecraft Forge Development Environment
This one is going to be a pain. We need the following:

* Downgrade Gradle enough to work with Forge *and* eliminate any bugs we can from the build process that we were finding before (the Kotlin Continuations bug)
* Install and configure UMC
* Make a basic mod class that can be loaded from the development environment

Baughn:
Fixes #16


Baughn:
The code quality failure can't be helped, I suppose.

PR #42, opened 2020-04-18
Add a Falstad importer

Grissess:
This one is not as fully featured, nor is it particularly robust
(handling fissing component errors gracefully is presently deferred to
upstream code) but it gets the job done, at least with resistive
circuits. Reactive elements remain untested.

Falstad's formatting prevents an easy way to access multipins, as
learned the hard way; some significant refactoring will likely need to
be made to achieve this. The abstraction was carefully chosen to allow
this in the future.

The code is untested except by example and the coverage probably
somewhat poor. Defer more testing until stabilization and probably merge
with an origin testing branch.

The stale old code is kept for reference right now, but should probably
be removed from any final releases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,9 @@ options.txt
 */classes
 */obj/
 
+# Graphviz Dot output files
+*.dot
+
 # IntelliJ
 .idea
 *.iml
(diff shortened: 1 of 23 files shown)


jrddunbr:
The Falstad.kt file needs more comments in the future, but I've done some comments and want to integrate this with some things in the `apps` folder, the rest looks good.

PR #43, opened 2020-04-19
Add materials and current cables

jrddunbr:


jrddunbr:
This branch is waiting for a thermal simulation

PR #47, opened 2020-04-21
Add documentation for uarchComputer

jrddunbr:
I added some documenation and also moved the tests out of the class they
were in in the testing suite. There still are some ambiguities about how
this operates since there is no microcode examples written, which also
means the main class is poorly covered in the testing infrastructure.


core/src/main/kotlin/org/eln2/compute/uarchComputer/IuOPROM.kt
```diff
@@ -1,7 +1,21 @@
...
+/**
+ * micro Operation ROM Interface - a method to store program data for the uarchComputer
+ */
 @ExperimentalUnsignedTypes
 interface IuOPROM {
+	/**
+	 * decode - Take integer, where to point to into the microcode rom to execute? - function pointer table
+	 * @param data Place to look up instruction
+	 * @return A decoded instruction
+	 */
```
jrddunbr, line 12: I asked @Caeleron about this but he should probably clarify it in the future.

PR #48, opened 2020-04-21
Reactive Component Bugfixes on the Falstad Importer

Grissess:
This PR contains some important changes to the MNA that should fix at least the reactive components--these regressions were discovered when the Falstad importer was used on the "basic LRC" circuit that [Falstad's simulator](http://www.falstad.com/circuit/) initially loads.


core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
```diff
@@ -130,15 +133,14 @@ class Falstad(val source: String) {
 		if(DEBUG) {
 			val repmap: MutableMap<PosSet, MutableSet<PosSet>> = mutableMapOf()
 			roots.values.forEach {
-				dprintln("F.<init>: r $it => ${it.representative}")
+				dprintln("F.<init>: r ${it} => ${it.representative}")
```
jrddunbr, line 136: need to revert this change

core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
```diff
@@ -130,15 +133,14 @@ class Falstad(val source: String) {
...
 				repmap.getOrPut(it.representative as PosSet, { mutableSetOf() }).add(it)
 			}
 
 			repmap.entries.forEach {
 				dprintln("F.<init>: R ${it.key}:")
 				it.value.forEach {
-					it2 ->
-					dprintln(" - $it2")
+					dprintln(" - $it")
```
jrddunbr, line 143: revert

core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
```diff
@@ -158,11 +160,10 @@ class Falstad(val source: String) {
...
 			(0 until ordered.size - 1).forEach {
-					it2 ->
-					ordered[it2].component.connect(
-					ordered[it2].pinidx,
-					ordered[it2+1].component,
-					ordered[it2+1].pinidx
+				ordered[it].component.connect(
+					ordered[it].pinidx,
+					ordered[it+1].component,
+					ordered[it+1].pinidx
```
jrddunbr, line 166: revert

core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
jrddunbr: This change is because the `it` variable shadowed the larger scope's variable, and I've seen that explode in fantastically nondeterministic ways at least in my experience.

core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
Grissess: The change here involves adding braces to a format string; this actually triggers a lint in IDEA ("Remove unnecessary braces"). Are you sure about this one?

Also, it's in the debug path, not at all hotpath or maintainability target.

core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
Grissess: It very much shouldn't explode; the lexical scoping here should be sensible. One does have to take care to note that the outer `it` is indeed inaccessible, or change the name via a named parameter.

If the change were a good idea, I'd certainly prefer a more descriptive name than `it2`.

core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
Grissess: idem, as above.

core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
jrddunbr: The git changes seem to be reverting the change, the braces are not needed (as IDEA says)

core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
jrddunbr: How about we just avoid the problem entirely and name it properly? I couldn't come up with a good name at the time.

core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt
Grissess: I'm dumb, I read the diff in the wrong order :)

PR #56, opened 2020-05-10
Draft client-server protocol for game to core communication

Baughn:


proto/src/logcollector.proto
```diff
@@ -0,0 +1,52 @@
...
+	// Included mainly for stability/crash analysis.
+	message OtherMod {
+		string name = 1;
+		Version version = 2;
+	}
+	repeated OtherMod mods = 1;
+	// Included for performance analysis.
+	message Circuit {
+		int32 matrix_size = 1;
+		int32 avg_time_usec = 2;
```
jrddunbr, line 40: Average time is not nearly as useful I don't think. In the least, average and maximum, but I would prefer to split it between matrix time, thevenin and norton process time, and other tick time. We don't currently have power sources nor intersystem, so we can't quite see where this will come into play, or if it will be relevant in this section or elsewhere in a protobuff - but I would value the input nonetheless to know how those times compare with different circuits.

Also, if I recall correctly, profiling the game while running in debug versions caused a pretty big performance hit, so we may want to be careful how we collect this data. Some benchmarks and testing in the mna code may be advised before making that change.

proto/src/logcollector.proto
```diff
@@ -0,0 +1,52 @@
...
+		repeated Component component = 3;
+	}
+	message Component {
+		string type = 1;
+		int32 count = 2;
+	}
```
jrddunbr, line 46: Would it be faster/more useful to just dump the entire connection matrix, with values for the connected components? That would make it so that we can replicate the state in testing and find things like singular matrices in commonly made circuits.

My guess is that would be undoubtedly more expensive.. but perhaps worth it at least as an opt-in for debugging someone's server if they want (eg, a "My world is having trouble, send current state report using a command"). Not sure.

proto/src/simulator/component.proto
```diff
@@ -0,0 +1,139 @@
...
+	// and for code de-duplication.
+	oneof type {
+		// Circuit primitives:
+		VoltageSource voltage_source = 10;
+		CurrentSource current_source = 11;
+		Resistor resistor = 12;
+		Inductor inductor = 13;
+		Capacitor capacitor = 14;
+		// Cables:
+		Resistor simple_cable = 20;
```
jrddunbr, line 38: Cables are really common and should probably be under ID 16 since it compresses the data in serialization.

proto/src/simulator/component.proto
```diff
@@ -0,0 +1,139 @@
...
+
+// In-world positioning.
+// This has no bearing on the simulation, unless perhaps if we implement a deeper thermal simulation.
+// Currently it's purely informational.
+message Location {
+	// Any of these may be missing if the question makes no sense, e.g. z-value for a 2D game.
+	sint32 x = 1;
+	sint32 y = 2;
+	sint32 z = 3;
+	sint32 world_id = 4;
```
jrddunbr, line 91: Some flexibility may be afforded if this is a string, but not particularly important.

proto/src/simulator/component.proto
```diff
@@ -0,0 +1,139 @@
...
+	float current_low = 3;
+	float current_high = 4;
+	float temperature_low = 5;  // Uh... for a chiller, maybe? In kelvin, of course.
+	float temperature_high = 6;
+}
+
+// In-world positioning.
+// This has no bearing on the simulation, unless perhaps if we implement a deeper thermal simulation.
+// Currently it's purely informational.
+message Location {
```
jrddunbr, line 86: This won't be able to handle SixNode. It doesn't store orientation data. My current guess is that if we were to use this, it would be inductor sharing, which is within a block. Also, storing as double may be advisable.

proto/src/simulator/component.proto
```diff
@@ -0,0 +1,139 @@
...
+
+// Pins: 2, one for each side.
+message Resistor {
+	float ohms = 1;
+}
+
+// Pins: 2, one for each side.
+message Inductor {
+	float inductance = 1;
+}
```
jrddunbr, line 112: Stores energy, may need a field for that.

proto/src/simulator/component.proto
```diff
@@ -0,0 +1,139 @@
...
+
+// Pins: 2, one for each side.
+message Capacitor {
+	float capacitance = 1;
+	// If set, fire a watchdog if U(P0) > U(P1).
+	bool polar = 2;
+}
```
jrddunbr, line 119: Stores energy, may need a field for that.

proto/src/simulator/simulator.proto
```diff
@@ -0,0 +1,116 @@
...
+	int64 unused_component_id = 3;
+}
+
+message TimestepRequest {
+	// Deliberately does not pass the timestep length.
+	// That's defined in a flag, and never changes afterwards.
+	//
+	// In practice this is a list of commands to be executed by the server.
+	repeated Command command = 2;
+}
```
jrddunbr, line 107: Vintage Story allows for changing the tick speed while the game is running IIRC. We should probably support it (to some degree)

proto/src/simulator/component.proto
jrddunbr: Also, is there really a reason to differentiate the cable from a standard resistor, since we would likely use line for both?

proto/src/logcollector.proto
```diff
@@ -0,0 +1,52 @@
...
+
+// Everything in this file is still in flux. It will be locked down prior to first alpha release.
+service LogService {
+	rpc Log(LogEntry) returns (LogResponse);
+}
+
+message LogEntry {
+	// Fields that identify the client:
+	Version minecraft = 1;
+	Version forge = 2;
```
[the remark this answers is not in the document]
Baughn: Done. Note that all fields are always optional, barring the use of a validation library... I couldn't find one that supports all the languages in question.

proto/src/logcollector.proto
Baughn: Well, it doesn't have to be profiled all the time. Statistical profilers can be made to have arbitrarily low performance impact.

I've split it out as suggested.

proto/src/logcollector.proto
Baughn: Maybe, but I have no clue how that looks. I'll have to request you provide the code for that one.

proto/src/logcollector.proto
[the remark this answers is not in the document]
jrddunbr: Alrighty

proto/src/simulator/component.proto
Baughn: TBH, that was supposed to be an example of the 'default to defining new components' rule. I later found a better one. For now I'll just remove it, but if cables end up doing *anything* different than resistors they should be a separate component type.

proto/src/simulator/component.proto
Baughn: We can always change it in the future. I'd rather have performance rather than avoid a five-minute change sometime later.

proto/src/simulator/component.proto
Baughn: That's a fair point, but OTOH it's meant to be advisory; it has no function inside the server.

I don't know how sixnode will end up looking here, given the SCM system I'm aiming for. Will add once that is clear, so probably much later.

proto/src/simulator/component.proto
Baughn: Done.

proto/src/simulator/component.proto
Baughn: Done.

proto/src/simulator/simulator.proto
Baughn: Hmm. I'm not familiar with VS, though. Can I ask you to add a suitable command after this is submitted?

proto/src/simulator/component.proto
jrddunbr: I can't currently see why cables would be any different from resistors as far as the MNA is concerned. In the thermal sim, yes, there will be a difference or two, but our cables are not going to become thermistors (resistors that change resistance as a result of temperature) because that would be prohibitively complicated to process and cause many unnecessary matrix recalcs on a per-thermal-tick level. Also, cables will be of a floating point length due to poles and other non-block cable types. Plus, I intend to introduce materials with different resistances and such.. but I don't think any of that matters to the MNA and as such doesn't need to be sent - it only needs to know about the resistance I think.

proto/src/logcollector.proto
jrddunbr: True, having the profiler running 24/7 is probably where the performance hit was. Just something to consider, I guess.

proto/src/logcollector.proto
```diff
@@ -0,0 +1,61 @@
...
+		Timing avg = 2;
+		Timing max = 3;
+		repeated Component component = 4;
+	}
+	message Component {
+		string type = 1;
+		int32 count = 2;
+	}
+	repeated Circuit networks = 2;
+	// What else is useful to collect stats on?
```
jrddunbr, line 57: User language?

proto/src/simulator/simulator.proto
jrddunbr: From the Vintage Story documentation:

> `/serverconfig tickrate [10-100]`
> Shows or sets the server tick rate. A higher tickrate makes the server more responsive. A lower tick rate has the opposite effect.

Baughn: Added a generic mechanism for flag updates.

Forge does also let you change mod settings, so that's fine.

PR #57, opened 2020-05-11
Document MNA

Grissess:
Pretty much what it says on the tin. Hopefully this can guide our wayward souls doing integration and support :)

### DIFF
--- a/core/src/main/kotlin/org/eln2/sim/electrical/mna/IDetail.kt
+++ b/core/src/main/kotlin/org/eln2/sim/electrical/mna/IDetail.kt
@@ -5,7 +5,11 @@ package org.eln2.sim.electrical.mna
  */
 interface IDetail {
 	/**
-	 * Produce a brief, debug-friendly description of this object.
+	 * Return a brief, debug-friendly description of this object.
+	 *
+	 * This is allowed to construct an object; the caller should not assume this will not result in allocations (as opposed to, e.g., a static string). Do not use this in a hotpath (at least in production).
+	 *
+	 * The semantics allow this operation to result in a different String value under various circumstances; it is not safe to assume the value is constant for any given object.
 	 */
 	fun detail(): String
 }
--- a/core/src/main/kotlin/org/eln2/sim/electrical/mna/IDetail.kt
+++ b/core/src/main/kotlin/org/eln2/sim/electrical/mna/IDetail.kt
@@ -1,5 +1,11 @@
 package org.eln2.sim.electrical.mna
 
+/**
+ * An interface for producing a brief description useful for debug annotation.
+ */
 interface IDetail {
+	/**
+	 * Produce a brief, debug-friendly description of this object.
+	 */
 	fun detail(): String
 }
--- a/core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
+++ b/core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
@@ -14,14 +14,12 @@ open class Node(var circuit: Circuit) : IDetail {
 	 * The potential of this node, in Volts, relative to ground (as defined by the [Circuit]); an output of the simulation.
 	 */
 	open var potential: Double = 0.0
+		internal set
 	/**
 	 * The index of this node into the [Circuit]'s matrices and vectors.
 	 */
 	open var index: Int = -1  // Assigned by Circuit
-	/**
-	 * True if this node is ground (defined to have zero potential).
-	 */
-	open val isGround = false
+		internal set
 	var name = "node"
 
 	override fun detail(): String {
@@ -44,8 +42,8 @@ open class Node(var circuit: Circuit) : IDetail {
 
 /**
  * A Node subclass for representing [Circuit.ground], with a higher [mergePrecedence], always [potential] 0V and [index] -1 (not assigned).
- * 
- * You can detect this via [isGround], which is true for these instances.
+ *
+ * The current recommended test for this special case is `node is GroundNode`.
  */
 class GroundNode(circuit: Circuit) : Node(circuit) {
 	override var potential: Double
@@ -55,8 +53,6 @@ class GroundNode(circuit: Circuit) : Node(circuit) {
 	override var index: Int
 		get() = -1
 		set(value) {}
-
-	override val isGround = true
 
 	override fun mergePrecedence(other: Node): Int = 100
 }
--- a/core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
+++ b/core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
@@ -15,12 +15,40 @@ open class Node(var circuit: Circuit) : IDetail {
 	 */
 	open var potential: Double = 0.0
 		internal set
+
 	/**
 	 * The index of this node into the [Circuit]'s matrices and vectors.
 	 */
 	open var index: Int = -1  // Assigned by Circuit
 		internal set
+
+	/**
+	 * True if this node is ground (defined to be 0V).
+	 */
+	open val isGround = false
+
+	/**
+	 * A name for this node, set by [named] (with a default usually assigned by the class).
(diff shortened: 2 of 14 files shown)


core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
```diff
@@ -1,18 +1,34 @@
...
+import org.eln2.space.Set
 
+/**
+ * A "node", in MNA; the non-resistive connections between [Component]s.
+ *
+ * Aside from identifying [Component]s' connections, the Nodes' potentials (relative to [Circuit.ground]) are computed as a result of the MNA algorithm.
+ */
 open class Node(var circuit: Circuit) : IDetail {
+	/**
+	 * The potential of this node, in Volts, relative to ground (as defined by the [Circuit]); an output of the simulation.
```
Baughn, line 14: Instead of trying to mark it as 'read-only' through documentation, you can make it private and export a public getter.

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
```diff
@@ -1,18 +1,34 @@
...
 	open var potential: Double = 0.0
+	/**
+	 * The index of this node into the [Circuit]'s matrices and vectors.
+	 */
 	open var index: Int = -1  // Assigned by Circuit
+	/**
+	 * True if this node is ground (defined to have zero potential).
+	 */
 	open val isGround = false
 	var name = "node"
```
Baughn, line 25: Is this really something that can change after construction?

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
```diff
@@ -1,18 +1,34 @@
...
 
 	override fun detail(): String {
 		return "[node val: $potential]"
 	}
 
-	/* Determine which node should prevail when two are merged in a Circuit.
+	/** Determine which node should prevail when two are merged in a Circuit.
 
 	   This is mostly so subclasses of Node (if any) can maintain their existence when merged. The Node returning the
```
Baughn, line 33: Hmm. How likely are such subclasses to exist?

Asking because there's a hefty performance penalty to not making the class sealed.

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
```diff
@@ -1,18 +1,34 @@
 package org.eln2.sim.electrical.mna
 
 import org.eln2.debug.dprintln
+import org.eln2.sim.electrical.mna.component.*
+import org.eln2.space.Set
 
+/**
+ * A "node", in MNA; the non-resistive connections between [Component]s.
```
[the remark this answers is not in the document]
Grissess: Having just found this after the thread, I suppose this is resolved, but for posterity:

A `Node` isn't a Pin _per se_--`PinRef` over in Falstad is much closer to that description. In particular, Nodes change identity when Components are connected, and a Node is not intended to be unique amongst all Components (whereas Pins are usually interpreted that way).

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
Grissess: It's a val, so I assume not? :)

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
Grissess: `GroundNode` is the big one--it takes preference in all cases. That's the _only_ one right now, though.

Work here mimics what I had to do to get the StationaryShaftNetwork working :)

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
Grissess: It gets set by `Circuit`, however, which is not in the inheritance chain. Not sure if `friend`s (as in C++) are a thing... :)

core/src/main/kotlin/org/eln2/sim/electrical/mna/IDetail.kt
```diff
@@ -1,5 +1,11 @@
 package org.eln2.sim.electrical.mna
 
+/**
+ * An interface for producing a brief description useful for debug annotation.
+ */
 interface IDetail {
+	/**
+	 * Produce a brief, debug-friendly description of this object.
```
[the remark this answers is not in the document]
Grissess: I guess I was thinking of a more active sense: "construct", "create"; a caller shouldn't assume this won't result in allocations. I'll cede it, though :)

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
Baughn: No, the name.

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
Baughn: You want `internal`, then.

Grissess: Throwing an `internal set` on the end of it should be good, then? I'll get to that with the next commit...

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
Grissess: The name `isGround` is problematic?

It's somewhat a shortcut to `node is GroundNode`, I suppose; I guess it could be replaced there if that test is faster. This current design affords some unneeded flexibility, now that the API is heading toward stable...

Grissess:
Ready for review; I think I checked all the boxes on this one.

I'm about to start a feature branch off of this one, since there were some subtle but important changes to the actual code, unlike the name of the branch would imply.

core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt
Baughn: Er.

What I mean is, can the value of `var name` actually change? Or should it be a val?

Grissess: ...oh, the field `name`... GitHub's UI for annotating comments leaves some to be desired. I, uh, deleted isGround in the newest tip anyway; d'ya think that's something I should reintroduce?

The name is... a good question! It probably should be a val, and honestly might even best be static. It was a part of the code that I believe I inherited from @jrddunbr and didn't touch; I don't know what his plan was for it. In particular, it's too vague for me to tell myself...

jrddunbr: I'm not convinced @Baughn is aware of how to tag the lines properly. I have the same issue; we should compare notes.

The old code had a "name" variable in the SubSystem implementation. This was handy because 2hen you dumped out a circuit, you had an inkling of an idea of what it was by the name. I'd like to expand on this so that components as part of a particular device (macerator, cable, battery,etc) are visible since they will be an abstraction layer above the MNA.

We could go about it a different way, I just found having a name is nice. I'd imagine things like "battery xxyyzz positive electrical load" (hopefully we rename the electrical load component by the way).

Grissess: okay, so name is a variable intended by use for a higher layer; it can definitely stay for that purpose :)

Should I revert the "removing `isGround`" change?

jrddunbr: Copying intents from Discord: yes, revert the change and apply changes for the `name` field.

jrddunbr:
I'm going to allow this to be pulled in as it has hit all of the major points, but we should increase test coverage soon.

PR #58, opened 2020-05-12
Add MultiMaps

Grissess:
MultiMaps relax the restriction that every K can associate with at most
one V; in general, a query for K can result in any set of V, including
empty ones.

These were already used in a few areas of code, refactored to consider
them.

### DIFF
--- a/core/src/main/kotlin/org/eln2/data/MultiMap.kt
+++ b/core/src/main/kotlin/org/eln2/data/MultiMap.kt
@@ -60,6 +60,11 @@ interface MultiMap<K, V> {
 	val entriesSize: Int get() = entries.map { 1 }.sum()
 }
 
+/**
+ * An implementor of [MultiMap] using the Kotlin stdlib.
+ *
+ * This class is defined here using a Map<K, Set<V>> idiom which is compatible with the Kotlin stdlib. Other, more performant implementations exist, and can be swapped in here if needed.
+ */
 class SetMapMultiMap<K, V>(iter: Iterator<Pair<K, V>>): MultiMap<K, V> {
 	/**
 	 * Internally implements the MultiMap interface as a Map<K, Set<V>>.
--- a/core/src/main/kotlin/org/eln2/data/MultiMap.kt
+++ b/core/src/main/kotlin/org/eln2/data/MultiMap.kt
@@ -1,0 +1,192 @@
+package org.eln2.data
+
+/**
+ * A helper class for representing a mapping as a Map.Entry.
+ */
+data class MultiMapEntry<K, V>(override val key: K, override val value: V): Map.Entry<K, V>
+
+/**
+ * A MultiMap: a data structure that maps one key (K) to zero or more values (V).
+ *
+ * This is in contrast to the usual Map, which maps one key to at most one value.
+ *
+ */
+interface MultiMap<K, V> {
+	/**
+	 * Tests whether a key [k] is in the MultiMap.
+	 *
+	 * This is true if k is associated with at least one value.
+	 */
+	operator fun contains(k: K): Boolean = get(k).isNotEmpty()
+
+	/**
+	 * Get the set of all values associated with key [k].
+	 */
+	operator fun get(k: K): Set<V>
+
+	/**
+	 * An iterator over every Entry<K, Set<V>>, of size keySize.
+	 */
+	val keyMapping: Iterable<Map.Entry<K, Set<V>>>
+	/**
+	 * The number of keys, value sets, and keyMapping entries in this MultiMap.
+	 */
+	val keyMappingSize: Int get() = keyMapping.map { 1 }.sum()
+	/**
+	 * The set of all keys.
+	 */
+	val keys: Set<K> get() = keyMapping.map {(k, _) -> k }.toSet()
+	/**
+	 * The set of all sets of values.
+	 * 
+	 * This will have the same size as keySize.
+	 */
+	val valueSets: Collection<Set<V>> get() = keyMapping.map {(_, vs) -> vs}
+	/**
+	 * The unique values in this MultiMap.
+	 */
+	val uniqueValues: Set<V>
+	/**
+	 * The number of unique values in this MultiMap.
+	 */
+	val uniqueValuesSize: Int get() = uniqueValues.size
+	/**
+	 * An iterator over every Entry<K, V>, where K may be repeated for each value.
+	 */
+	val entries: Iterable<Map.Entry<K, V>>
+	/**
+	 * The number of pairings in this MultiMap.
+	 */
+	val entriesSize: Int get() = entries.map { 1 }.sum()
+}
+
+class SetMapMultiMap<K, V>(iter: Iterator<Pair<K, V>>): MultiMap<K, V> {
+	/**
+	 * Internally implements the MultiMap interface as a Map<K, Set<V>>.
+	 */
+	val map: Map<K, Set<V>> = with(mutableMapOf<K, MutableSet<V>>()) {
+		iter.forEach { (k, v) -> getOrElse(k, { mutableSetOf() }).add(v) }
+		entries.associate { (k, v) -> Pair(k, v.toSet()) }
+	}.toMap()
+
+	constructor(): this(emptyList<Pair<K, V>>().iterator())
+
+	override fun get(k: K): Set<V> = map.get(k) ?: emptySet()
+
+	override val keyMapping: Iterable<Map.Entry<K, Set<V>>> = map.entries
+	override val keyMappingSize: Int = map.size
+	override val keys: Set<K> get() = map.keys
+	override val valueSets: Collection<Set<V>> get() = map.values
+	override val uniqueValues: Set<V> get() = map.values.fold(emptySet()) { next, acc -> acc.union(next) }
+	override val entries: Iterable<Map.Entry<K, V>> get() = map.entries.flatMap { (k, vs) -> vs.map { MultiMapEntry(k, it) }}
+}
+
+/**
+ * A Mutable version of the [MultiMap] interface.
+ */
+interface MutableMultiMap<K, V>: MultiMap<K, V> {
+	/**
+	 * Get the set of values for the key [k].
+	 *
+	 * Note that the return is mutable; one can add or remove entries from this map by mutating it.
+	 */
+	override operator fun get(k: K): MutableSet<V>
+
+	/**
+	 * Add an association from [k] to [v] (in addition to any other associations involving [k] or [v]).
+	 */
+	operator fun set(k: K, v: V) = get(k).add(v)
+
+	/**
+	 * Clear all values associated with the key [k].
+	 */
+	fun clear(k: K)
+
+	/**
+	 * Remove an association between [k] and [v].
+	 *
+	 * If the map no longer contains [k], [clear] is invoked (as it usually performs some implementation-specific cleanup).
+	 */
+	fun remove(k: K, v: V) {
+		val s = get(k)
+		s.remove(v)
+		if (s.isEmpty()) clear(k)
+	}
+}
+
+/**
+ * An implementor of [MutableMultiMap] using the Kotlin stdlib.
+ *
+ * This class is defined here using a MutableMap<K, MutableSet<V>> idiom which is compatible with the Kotlin stdlib. Other, more performant implementations exist, and can be swapped in here if needed.
+ */
+class MutableSetMapMultiMap<K, V>: MutableMultiMap<K, V> {
+	val map: MutableMap<K, MutableSet<V>> = mutableMapOf()
+
+	constructor(iter: Iterator<Pair<K, V>>) {
+		iter.forEach { (k, v) -> set(k, v) }
+	}
+
+	override operator fun get(k: K): MutableSet<V> = map.getOrElse(k, { mutableSetOf() })
+
+	override fun clear(k: K) {
+		map.remove(k)
+	}
+
+	override val keyMapping: Iterable<Map.Entry<K, Set<V>>> = map.entries
+	override val keyMappingSize: Int = map.size
+	override val keys get() = map.keys
+	override val valueSets: Collection<Set<V>> = map.values
+	override val uniqueValues: Set<V> get() = map.values.fold(emptySet()) { next, acc -> acc.union(next.toSet()) }
+	override val entries: Iterable<Map.Entry<K, V>> get() = map.entries.flatMap { (k, vs) -> vs.map { MultiMapEntry(k, it) }}
+}
+
+/**
+ * Create an empty MultiMap.
+ * 
+ * The choice of map is an implementation detail, but chosen to be a good compromise for general use.
+ */
+fun<K, V> emptyMultiMap(): MultiMap<K, V> = SetMapMultiMap<K, V>()
+
+/**
+ * Creates a MultiMap over the given pairs (`a to b`).
+ *
+ * The choice of map is an implementation detail, but chosen to be a good compromise for general use.
+ */
+fun<K, V> multiMapOf(vararg pairs: Pair<K, V>): MultiMap<K, V> = SetMapMultiMap(pairs.iterator())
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
...
+
+	/**
+	 * Get the set of all values associated with key [k].
+	 */
+	operator fun get(k: K): Set<V>
+
+	/**
+	 * An iterator over every Entry<K, Set<V>>, of size keySize.
+	 */
+	val keyMapping: Iterable<Map.Entry<K, Set<V>>>
```
Baughn, line 30: Nit: Missing linebreak.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
...
+	/**
+	 * The number of keys, value sets, and keyMapping entries in this MultiMap.
+	 */
+	val keyMappingSize: Int get() = keyMapping.map { 1 }.sum()
```
Baughn, line 34: Could you use .count() instead?

I know it's an iterable, but the underlying data structure isn't.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
...
+	/**
+	 * The set of all keys.
+	 */
+	val keys: Set<K> get() = keyMapping.map {(k, _) -> k }.toSet()
```
Baughn, line 38: Doesn't this already have .keys()?

Also, newline.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
...
+	/**
+	 * The set of all sets of values.
+	 * 
+	 * This will have the same size as keySize.
+	 */
+	val valueSets: Collection<Set<V>> get() = keyMapping.map {(_, vs) -> vs}
+	/**
```
Baughn, line 45: Newline.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
...
+	 * The unique values in this MultiMap.
+	 */
+	val uniqueValues: Set<V>
+	/**
```
Baughn, line 49: Newline.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
...
+	 * The number of unique values in this MultiMap.
+	 */
+	val uniqueValuesSize: Int get() = uniqueValues.size
+	/**
```
Baughn, line 53: Newline.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
...
+	 * An iterator over every Entry<K, V>, where K may be repeated for each value.
+	 */
+	val entries: Iterable<Map.Entry<K, V>>
+	/**
```
Baughn, line 57: Newline.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
...
+	 * The number of pairings in this MultiMap.
+	 */
```
Baughn, line 59: Probably want to add `O(n)` to the docs.

In fact, ideally add algorithmic complexity to every method. It's not always obvious; without reading the implementation, I'd assume a getter for a single int is always O(1).

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
+package org.eln2.data
```
Baughn, line 1: Please add tests for all of this.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * A helper class for representing a mapping as a Map.Entry.
+ */
+data class MultiMapEntry<K, V>(override val key: K, override val value: V): Map.Entry<K, V>
```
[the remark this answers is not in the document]
Grissess: Map.Entry is an interface, unfortunately; that's really all there is to this :(

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Grissess: Didn't know about count! Will do once I get back to an IDE :)

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Grissess: You mean "no blank line"? I sometimes do that for things that are "logically grouped". I can add it if you think it's worth it...

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Grissess: You do have a point... the current implementations override this to be O(1) anyway. This is just the _ultimate_ fallback--if you can do better than this, you should.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Baughn: I sometimes use vim for editing, and its notion of a 'paragraph' is just anything that's separated by blank lines. So I write code such that this is useful.

I'd prefer something more visible than a blank line to separate logical sections.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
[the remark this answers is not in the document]
Baughn: Noted.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Baughn: If  _every_ implementation does better, then there's no need for the base case; it becomes nothing more than a potential (performance) bug in the future.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Grissess: You have a point; that'll be the next commit, then.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Grissess: I'm still of the mind of reducing implementation burden for the future, but in this case:
1. The implementation here is likely to be a footgun for a lazy implementor, and
2. This is the core codebase where we don't expect a lot of naive developers to be poking around.

I'll leave the current implementation in a comment as a suggestion, but you're right in that it might not be best as a _fallback_.

Grissess:
Aaaalright. Tests were added, cleanups were done, I think this is ready for another look-over. I'm definitely not happy with the structure of that test (and how much of it is _abject copying_), but--until I find out more about JUnit--it _does_ stand and work.

Score one for the testing framework: I was using `getOrElse` when I meant to use `getOrPut`. That bug was stomped out nice and early.

(Propose ready for review.)

Grissess:
(just rebased onto the new master, sorry about that; ready for review again.)

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,231 @@
...
+ *
+ * This class is defined here using a Map<K, Set<V>> idiom which is compatible with the Kotlin stdlib. Other, more performant implementations exist, and can be swapped in here if needed.
+ */
+class SetMapMultiMap<K, V>(iter: Iterator<Pair<K, V>>): MultiMap<K, V> {
+	/**
+	 * Internally implements the MultiMap interface as a Map<K, Set<V>>.
+	 */
+	val map: Map<K, Set<V>> = with(mutableMapOf<K, MutableSet<V>>()) {
+		iter.forEach { (k, v) -> getOrPut(k, { mutableSetOf() }).add(v) }
+		println("SMMM.map: $this")
```
Baughn, line 99: Forgot to remove this?

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,231 @@
...
+		entries.associate { (k, v) -> Pair(k, v.toSet()) }
```
Baughn, line 100: This could easily be extremely inefficient.

If the class exists as a comparison for tests/documentation/fuzz comparisons, that's fine, but in that case it should be annotated with @TestOnly. If it's currently in use, then... well, I suppose that's fine for now; just add a "TODO: Add performance tests".

core/src/main/kotlin/org/eln2/data/MultiMap.kt
```diff
@@ -0,0 +1,231 @@
...
+		if (s.isEmpty()) clear(k)
+	}
+}
+
+/**
+ * An implementor of [MutableMultiMap] using the Kotlin stdlib.
+ *
+ * This class is defined here using a MutableMap<K, MutableSet<V>> idiom which is compatible with the Kotlin stdlib. Other, more performant implementations exist, and can be swapped in here if needed.
+ */
+class MutableSetMapMultiMap<K, V>: MutableMultiMap<K, V> {
```
Baughn, line 157: If you want to reduce the duplication, I think you can implement SetMapMultiMap in terms of MutableSetMapMultiMap, restricting the interface appropriately. (And wow is that a mouthful.)

Personally, I'm fine with this.

core/src/test/kotlin/org/eln2/data/SetMapMultiMapTest.kt
```diff
@@ -0,0 +1,286 @@
+package org.eln2.data
+
+import org.junit.jupiter.api.Assertions.*
+import org.junit.jupiter.api.Test
+
+// FIXME: abstract this so that the Mutable version below isn't LITERALLY an unmaintainable copy of this
```
Baughn, line 6: The way to do this is to make both interfaces inherit from a super-interface that just defines the common operations, but...

I'm not sure exactly how to do that.

core/src/test/kotlin/org/eln2/data/SetMapMultiMapTest.kt
```diff
@@ -0,0 +1,286 @@
...
+
+		assert(TEST_NONKEY !in map)
+		assert(map[TEST_NONKEY].isEmpty())
+	}
+
+	@Test
+	fun multiMapConsistency() {
+		val map = makeImplementation(TEST_MULTIMAP)
+
+		println("multimap keymappings ${map.keyMapping} entries: ${map.entries} size ${map.entriesSize}")
```
Baughn, line 69: Tests shouldn't print anything except if they fail.

core/src/test/kotlin/org/eln2/data/SetMapMultiMapTest.kt
```diff
@@ -0,0 +1,286 @@
...
+		assert(TEST_NONKEY !in map)
+		assert(map[TEST_NONKEY].isEmpty())
+	}
+}
+
+internal class MutableSetMapMultiMapTest {
+	// ------------- START PASTING THE BODY OF THE CLASS ABOVE HERE --------------
+
+	companion object {
+		val TEST_MAP: Map<Int, Int> = mapOf(1 to 2, 2 to 4, 4 to 8, 8 to 16, 16 to 32)
```
Baughn, line 96: You can definitely put this test-data in a global val.

core/src/test/kotlin/org/eln2/data/SetMapMultiMapTest.kt
Grissess: You heard my tribulations in real time when I was asking what would happen if you made an "abstract test driver class"; my preferred structure would have been an abstract method to make the instance under test, and the code superficially resembles this right now :)

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Grissess: Yep.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Grissess: I've discovered it is much a similar problem as what occurs with Rust, and one of the motivations for their `MaybeInit<T>` union (that replaces a bunch of unsafe raw pointer scariness and `mem::uninitialized()`): even though the data structure is supposed to be immutable, _initializing it_ somehow tends to cause mutations (if you don't dig too deeply into the details). I have no idea how to fix this specifically in Kotlin, so I suspect I'll end up having to go for the "TODO" option.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Grissess: It's secretly Java, isn't everything supposed to be a mouthful? :)

I... might take you up on that, actually; it's no issue having the underlying mutable structure, I suppose, if you don't expose it. (This assumes "immutable casts" are cheap, like `MutableSet::toSet`, which, incidentally, for this hypothetical implementation, the `MutableSetMapMultiMap::toSetMapMultiMap()` would be--it returns `this`.)

core/src/test/kotlin/org/eln2/data/SetMapMultiMapTest.kt
Grissess: Sorry, some left-over debug prints. I feel like that could guide someone hunting a regression, though; mind if I leave it there but commented as a "known good debug tool"? (I don't imagine leaving commented-out code in _test cases_ is the most evil thing...)

core/src/test/kotlin/org/eln2/data/SetMapMultiMapTest.kt
Grissess: Bearing in mind this was all supposed to be in one (abstract) class, by the time it came down to yank-pasting the test classes, I was already frustrated enough that this didn't occur to me :)

It will be done.

core/src/test/kotlin/org/eln2/data/SetMapMultiMapTest.kt
Baughn: You could also use a try-finally, I think.

Grissess:
forgot to mention: ready for next review :)

The println is a dprintln, now; it's not terrible using the debug infrastructure in tests, right?

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Baughn: I don't think you addressed this.

Grissess: You're right, I missed this :)

`keyMapping: Iterable<Map.Entry<K, Set<V>>>` which... doesn't have `keys` on it, if that's what you're asking; you _could construct a `Map` out of it proper and take its keys, but I was avoiding round-tripping through another object for that, especially since the `Map`-dependent code is in the concrete implementation.

At least, I think that's what you're asking? Let me know if you meant something else :)

jrddunbr:
I haven't determined why, but Codacy reports some of these files have issues and they don't seem to exist when I actually drill down, but in other files where they do exist, it doesn't show up on the main page. I'm going to propose we use a different code quality tool that actually works.

core/src/main/kotlin/org/eln2/data/MultiMap.kt
Baughn: Well, keyMapping is constructing an iterable from a concrete map which does have .keys(). But let's leave that as a todo for the future.

PR #59, opened 2020-05-13
Updated documentation and removed dead codebases

jrddunbr:
I added some more details to the contributing document, modified the
list of contributors to accurately reflect the current state of things,
and made the readme more concise.


jrddunbr:
The labeler is drunk.

jrddunbr:
Requesting review of workflow document as you proposed previously. I expect within 2-3 weeks that we move development to the `develop` branch, but I do not really want to move it before we have a beta release that would have:

* analytics
* user opt-in to development release
* any kind of MNA tie-in. If you can place a resistor and a voltage source and it sim's, fantastic!

CONTRIBUTING.md
```diff
@@ -2,6 +2,52 @@
...
 
-Feel free to make pull requests or issues, but please note that we are currently focused on making the mod, not adding more content.
+Feel free to make pull requests or issues, but please note that we are currently focused on making the mod functional, before adding more content.
 
-People who make pull requests should be familiar with the mod licenses; if you make a PR, you agree to the license.
+## PR Guidelines
+
+* Keep PR's to one topic, if possible. This permits us to use version control to find software regressions more easily.
+* Avoid mass operations such as folder moves and restructures next to logical changes in the code.
+* Use the PR message to describe changes; use markdown check marks to show intended progress towards a goal.
```
Baughn, line 11: PR messages disappear after it's merged. We also need good commit messages.

Link to https://chris.beams.io/posts/git-commit/?

CONTRIBUTING.md
```diff
@@ -2,6 +2,52 @@
 
 Contact us on our Discord at [https://discord.gg/YjK2JAD](https://discord.gg/YjK2JAD)
 
-Feel free to make pull requests or issues, but please note that we are currently focused on making the mod, not adding more content.
+Feel free to make pull requests or issues, but please note that we are currently focused on making the mod functional, before adding more content.
 
-People who make pull requests should be familiar with the mod licenses; if you make a PR, you agree to the license.
+## PR Guidelines
+
+* Keep PR's to one topic, if possible. This permits us to use version control to find software regressions more easily.
```
Baughn, line 9: It's far more important to keep *commits* to one topic.

CONTRIBUTING.md
```diff
@@ -2,6 +2,52 @@
...
+* Avoid mass operations such as folder moves and restructures next to logical changes in the code.
+* Use the PR message to describe changes; use markdown check marks to show intended progress towards a goal.
+* Keep PR up to date with the `master` branch to ease merges. In the future, once we have one, keep up to date with `develop` instead.
```
Baughn, line 12: I disagree with the second sentence. 'master' should be enough on its own; we can create release branches as needed.

CONTRIBUTING.md
```diff
@@ -2,6 +2,52 @@
...
+* PR's should come with their code documented and unit tested (if possible).
```
Baughn, line 13: We reserve the right to push back if we disagree on whether it's possible.

CONTRIBUTING.md
```diff
@@ -2,6 +2,52 @@
...
+* If you make a PR, you agree to the licenses of the project, and that the code and content you are contributing complies with the applicable license.
+
+## Branch Naming
+
+Please use concise names that accurately describe what you are doing. For example, `feature/add-variable-resistor` may describe some new Variable Resistor you are adding.
+
+The following naming convention is to be used, with the exception that currently `master` is being used as `develop` (see below comment).
+
+* `master` – Stable release branch that is known for production level stability and has hotfixes applied if that is not the case.
```
Baughn, line 22: That's describing a release branch.

CONTRIBUTING.md
```diff
@@ -2,6 +2,52 @@
...
+    * math: Math helpers
+    * parsers: Parsers and import/export tools
+    * sim: Electrical and Thermal simulators
+    * space: various data structures and algorithms
+* images: Images for the falstad importer when exporting circuit data
+* integration-mc1-15: Minecraft 1.15.2 Integration Code
+* integration-vs: Vintage Story Integration Code
+* logcollector: A rust statistics and (crash) logging server
+* proto: A set of [Google Protocol Buffers](https://developers.google.com/protocol-buffers) for communication with all of our applications (proto3)
+* shared: For code that may be shared amongst platforms (maybe)
```
Baughn, line 60: Honestly, I feel like this'll all end up in core.

CONTRIBUTING.md
jrddunbr: The reason I suggest PR messages is to keep track of *progress* on a branch. It's meant to be modified and not saved in source control.

For example, I may have this:

## Variable Resistor
- [x] MNA Code
- [ ] Block
- [ ] User Interface

and by the time they PR it, those are checked off and there are commits that have messages that are useful.

CONTRIBUTING.md
jrddunbr: True.

CONTRIBUTING.md
jrddunbr: I think that is implied here. Obviously, if we don't like what we see, we don't merge it.

Baughn: I think adding the words will make contributors think a bit harder before deciding they can't test it.

CONTRIBUTING.md
jrddunbr: Did you mean to select all of those lines or just `shared`? I was actually going to propose we removed `shared` and re-add it later, if ever needed. I originally came up with `shared` because I thought there was going to be a lot more flack to having most of the codebase not be in `integration`.

jrddunbr: Oh, also, in a seperate PR, we should remove the Vintage Story gradle project. I doubt there will ever be any Kotlin/Java in there.

jrddunbr:
I removed those labels because I view this stuff more as a meta topic, rather than being related to any particular part of the code. It's detecting that code has changed in the project repository in particular folders, but I would argue that conventions of how the code is written is not related to code changes in Core which is what that label is targeted at. I guess removing the folders for older Minecraft code makes sense though.

CONTRIBUTING.md
jrddunbr: Second point actually was in a PR I was about to submit, funny enough.

Baughn: Makes sense, but I still think encouraging good commit hygiene is good.

CONTRIBUTING.md
Baughn: I didn't select *any* lines. Github just does that on its own.

Also, concur.

Baughn:
> 
> 
> I removed those labels because I view this stuff more as a meta topic, rather than being related to any particular part of the code. It's detecting that code has changed in the project repository in particular folders, but I would argue that conventions of how the code is written is not related to code changes in Core which is what that label is targeted at. I guess removing the folders for older Minecraft code makes sense though.

Well, that means your PR is violating the 'single-topic' convention explained in this PR. 🧛

jrddunbr:
True, that's why I removed that and put it to "per commit". I realized that it was a bit too restrictive - for instance, adding a feature may require files in both core and integration, and there's no good reason to separate those, and we don't have a great mechanism to handle desyncronizations in the supported code between the two. Ideally, we would have some code that anticipates this and handles it gracefully if possible.

Also, at the point at which we are, those labels do make sense. I was more aiming that it seems to be adding labels where they didn't make sense; obviously the scope has changed of the PR.

jrddunbr:
Oh, I deleted my trash serialization code attempt, right. No, I guess the labeler is right here. Carry on..

CONTRIBUTING.md
jrddunbr: .. how does GitHub select random lines? It only allows you to select lines to comment on, or to add them to the review comment. I don't think it accepts review comments besides that which are un-associated to the code... maybe I'll play with this later just to confirm.

PR #61, opened 2020-05-17
Adds Multimeter item and client translations

jrddunbr:
I added a new multimeter item. It doesn't really do anything, but I
upscaled the texture and set up the language localization. According to
the forge documentation, you can set your own translation keys, and I
didn't see much of a convention, so I set it to
tooltip.modid.translationtype.itemorblock.

I also added the translation and tooltip to Flubber. Yay!


jrddunbr:
I'm just going to merge this. We can go back to this and revise it with commits, but it doesn't really do too much that's in need of review.

PR #62, opened 2020-05-18
Adds basic ore code to Minecraft

jrddunbr:
Added some basic copper ore to Minecraft. I was going to work on
ItemGroups to go with it but that turned into a dumpster fire quickly.

I also refactored the block loading in the main class to use a Kotlin
lambda function.

We have a list of other ores that we plan to add, but I do not have the
textures from my texture artist yet, those will be available in a few
days hopefully.

This also adds some code that does the actual population of the ores in
the world. I did not balance the copper at all, it was high enough to
make sure it worked.

When you break the ore block, it will drop a different item; this item
is not able to be placed again (because that doesn't make sense!).
I don't plan to allow Silk Touch to be able to pick up these blocks at
all, but I have not verified that.


integration-mc1-15/src/main/kotlin/org/eln2/ModBlocks.kt
```diff
@@ -1,11 +1,15 @@
...
+import org.eln2.blocks.OreBlock
+import org.eln2.utils.OreData
 
 /**
  * Blocks added here are automatically registered.
  */
 enum class ModBlocks(val block: Block) {
-	FLUBBER(FlubberBlock())
+	FLUBBER(FlubberBlock()),
+	ORE_NATIVE_COPPER(OreBlock(OreData(1.0f, "ore_native_copper", ModItems.ORE_CHUNKS_COPPER.items.defaultInstance)))
```
Baughn, line 14: Ideally, all blocks of the same 'type' will be variations of the same block using meta data. The total block ID count is still limited. At least, it was in 1.12; it's possible that 1.15 has changed something.

That's something I can refactor once we have >1 ore blocks, though.

integration-mc1-15/src/main/kotlin/org/eln2/ModItems.kt
```diff
@@ -1,11 +1,14 @@
...
 import org.eln2.items.MultimeterItem
+import org.eln2.items.OreChunks
 
 /**
  * Items added here are automatically registered.
  */
 enum class ModItems(val items: Item) {
-	MULTIMETER(MultimeterItem())
+	MULTIMETER(MultimeterItem()),
+	ORE_CHUNKS_COPPER(OreChunks())
```
Baughn, line 13: Bookshelf automatically creates an item for each registered block, unless overridden, which we don't currently. You shouldn't need to add one here, and if you do, you'll need to change the block registration to support passing in the item.

The way you're doing things here, you'll get a separate block-item (created by Bookshelf) from the item dropped by the ore when broken. That seems to be what you want, but it's worth a comment.

integration-mc1-15/src/main/kotlin/org/eln2/blocks/OreBlock.kt
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * OreBlock - a block of ore.
+ */
+class OreBlock(val oreType: OreData): Block(Properties.from(Blocks.IRON_ORE).hardnessAndResistance(oreType.hardness)) {
+	/**
+	 * getDrops override so that we drop the ore dust.
+	 *
+	 * Yes, it's deprecated, no, I don't care. I dare you to look at the supported method, if you can find it.
+	 * TODO: Stop using deprecated function
```
Baughn, line 18: Stupidly, "Deprecated" in Forge means "Don't call this". It doesn't mean "Don't override this."

integration-mc1-15/src/main/kotlin/org/eln2/items/OreChunks.kt
```diff
@@ -0,0 +1,19 @@
...
+import net.minecraft.item.ItemStack
+import net.minecraft.util.text.ITextComponent
+import net.minecraft.util.text.TranslationTextComponent
+import net.minecraft.world.World
+
+/**
+ * It's a chunk of ore! This is what you get when you break the ore block.
+ * Minecraft handing you a block that you can place makes ~0 sense.
+ */
+class OreChunks: Item(Properties().maxStackSize(64)) {
```
Baughn, line 14: Isn't 64 already the default?

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreData.kt
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * OreData
+ * Data about the ores you are trying to make.
+ *
+ * @param hardness How hard is the block (pickaxe type basically)
+ * @param name Name of the ore (eg, "native_copper")
+ * @param result The item to drop
+ */
+data class OreData(val hardness: Float, val name: String, val result: ItemStack)
```
Baughn, line 13: Can we use an enum instead (for hardness), just to reduce confusion? Hand, wood, stone, iron, diamond...

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreGen.kt
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Handles ore generation for Eln2
+ */
+class OreGen {
+	companion object {
+		/**
+		 * Configures ore generation in the world. Registers against Forge.
+		 */
+		fun setupOreGeneration() {
```
Baughn, line 22: This should definitely be configurable. Add a TODO, at least.

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreGen.kt
```diff
@@ -0,0 +1,71 @@
...
+			ForgeRegistries.BIOMES.filter {
+				it.category != Biome.Category.NETHER && it.category != Biome.Category.THEEND
+			}.forEach {
+				/*
+				Flubber generation
+
+				it.addFeature(
```
Baughn, line 29: ...would make a fun easter egg.

Baughn (changes requested):

integration-mc1-15/src/main/kotlin/org/eln2/ModBlocks.kt
jrddunbr: That sounds like the best approach.

integration-mc1-15/src/main/kotlin/org/eln2/items/OreChunks.kt
jrddunbr: It may be, but I didn't see any harm in specifying. Later though, I may base this around storing NBT and having mass to reduce the number of items you store.

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreData.kt
jrddunbr: Does Minecraft already have such an enum? I am unfamiliar with how hardness works, I thought it affected blast resistance mostly and also the level at which an iron pick is required.

Baughn: It probably has constants, but those don't make it through the decompiler.

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreGen.kt
jrddunbr: I intended to leave this as a function that doesn't need to be called in by the main class, ie, it's a registration method. This is the configurable part; you tell it what blocks, and what levels to put them at. Do you want to clutter the main class with that registration code?

Baughn: It isn't configurable _at the moment_, hence the TODO.

integration-mc1-15/src/main/kotlin/org/eln2/blocks/OreBlock.kt
jrddunbr: *sigh*. Do you know a better way to do this without making a mess of classes everywhere? And, the forge developers should really learn how to use annotations properly.

Baughn: No, no I don't. It's one reason I'm pushing for the internal API, to separate out the mess.

jrddunbr: I'm a fan of pulling this and leaving it a TODO. The code works and doesn't seem to explode my game. When we get an internal API, this would be a good spot to start perhaps.

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreData.kt
jrddunbr: Time to make an enum then. Any documentation you know of that documents what those hardness levels are?

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreGen.kt
jrddunbr: What isn't configurable about it exactly? You literally, in one statement, register what ores you want to be generated and how. I don't see how I could make this any more configurable besides adding a literal configuration option in Eln2.json (or whatever our config file will be).

Baughn: Well, yes, that's what I mean. It isn't user-configurable, and it would be easy to forget which things should be when we get around to adding that.

Hence the TODO.

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreData.kt
Baughn: Well, there's https://minecraft.gamepedia.com/Module:Hardness_values

integration-mc1-15/src/main/kotlin/org/eln2/ModItems.kt
jrddunbr: So I know there are item blocks, but the point of the code here is that ore chunks are *not* the original ore block. The problem with Minecraft `ItemBlock`'s is that you can place them again, eg, if you break an iron ore block, you can place it again. I don't want that, and later, we may even make it so that these piles of dust have NBT/internal data concerning what is actually in the particular sample of ore; eg, 30% iron (ii) oxide, 60% stone, 10% copper (ii) oxide... and convert that to a mass.

The TL;DR is that I don't think Bookshelf has the capability to do what I want around these blocks, and later when it drops items with data, it really won't help.

Baughn: What Bookshelf is doing right now is: Create an item which can be used to place a copper ore block, but which won't drop when you break it.

That's probably what you want. I'm just noting it's unexpected behaviour to anyone new to the codebase.

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreData.kt
jrddunbr: Actually, nevermind. Apparently, the block hardness affects both the tool and the mining speed, and many of the blocks in Minecraft have values all over the darned place. We could add constants that have particular minimum values for say an iron pickaxe, but there is no strict set of values they seem to be using.

https://minecraft.gamepedia.com/Breaking#Blocks_by_hardness

integration-mc1-15/src/main/kotlin/org/eln2/utils/OreGen.kt
jrddunbr: Todo added.

jrddunbr:
Fixed the above.

PR #64, opened 2020-05-18
SCAM Parser [WIP]

jrddunbr:
Basic thing with netlists. My code is terrible. WIP

### DIFF
--- a/core/src/main/kotlin/org/eln2/parsers/scam/Scam.kt
+++ b/core/src/main/kotlin/org/eln2/parsers/scam/Scam.kt
@@ -1,0 +1,67 @@
+package org.eln2.parsers.scam
+
+import org.eln2.parsers.scam.components.*
+import org.eln2.sim.electrical.mna.Circuit
+import org.eln2.sim.electrical.mna.component.Component
+import org.eln2.space.Set
+
+data class PinSet(val pos: String): Set()
+
+interface IComponentConstructor {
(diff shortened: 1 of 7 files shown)


jrddunbr:
If anyone wants to take a crack at this, feel free.

core/src/main/kotlin/org/eln2/parsers/scam/Scam.kt
```diff
@@ -0,0 +1,67 @@
...
+		val TYPE_CONSTRUCTORS: MutableMap<Char, IComponentConstructor> = mutableMapOf(
+			'R' to ResistorConstructor(),
+			'L' to InductorConstructor(),
+			'C' to CapacitorConstructor(),
+			'V' to VoltageSourceConstructor(),
+			'I' to CurrentSourceConstructor()
+		)
+
+		fun buildFromString(sourceLine: String, circuit: Circuit) {
+			val all = sourceLine.split("\n")
```
Baughn, line 21: Does looks like a manually written parser.

It's fine in this case, since it's small, but if we're going to have anything significantly more complicated then we should use https://copper-leaf.github.io/kudzu/ or something.

jrddunbr: This is absolutely a manually written parser. It will not be very complex.

PR #65, opened 2020-05-19
Material Properties [WIP]

jrddunbr:
It adds some basic material properties, but I'm sure there's more to be added, and
this could be restructured as well perhaps. But the enum is nice.


jrddunbr:
Cam and I were talking yesterday, and we're not sure how well these values will play with the matrix condition number, but hopefully we can give it a try and if it doesn't work, we can modify the numbers some to be better behaving and within reasonable limits.

jrddunbr:
I'm merging this because it seems like a good start and it's just going to fall behind at this point.

PR #70, opened 2020-05-22
Tests and improvements for org.eln2.math

Caeleron:


core/src/main/kotlin/org/eln2/math/IFunction.kt
```diff
@@ -11,5 +11,5 @@ interface IFunction {
 	 * @param x The input variable
 	 * @return The output variable
 	 */
-	fun getValue(x: Double): Double
+	fun getValue(x: Double): Double?
```
jrddunbr, line 14: `fun getValue(x: Double): Double?` should be `fun getValue(x: Double): Double`

core/src/main/kotlin/org/eln2/math/FunctionTable.kt
```diff
@@ -3,38 +3,51 @@ package org.eln2.math
...
+				}
+			}
+			lx > 1.0 -> { // Extrapolation to the right.
+				when (mode) {
+					ExtrapolationMode.ClosestValue -> point.last()
+					ExtrapolationMode.Linear -> point.last() + (point.last() - point[point.size - 2]) * point.last() * (lx - 1.0)
+				}
+			}
+			else -> { // Interpolation
+				val sx = lx.times(point.size - 1)
```
Grissess, line 42: just curious: why the non-operator `times` here?

jrddunbr: My guess is that when IDEA converted this file from Java to Kotlin that it thought to keep this? That's my best guess, but it's not necessary to keep this way.

PR #71, opened 2020-05-22
Component Graph

Grissess:
This is likely to be a useful abstraction in two very important places:

- The "Managed Circuit" API (to soon be coming after this) which abstracts the core MNA Circuit abstraction and relaxes the issues with removing components; and
- I posit, the thermal simulation, where each region is a simulation domain (this might not have such important performance benefits as the former, but nonetheless).

Note: this is built off of the `feature/multimap` branch because the tests depend on multimaps.

Todo:
- [ ] Improve test coverage
- [ ] Document
- [ ] Independently review API ergonomics (test implement CircuitManager off this tip?)


Grissess:
I'll take a look at this in the morning; the linter is likely to flip out about whitespace, but the algorithms seem solid. This isn't highly-optimized at the moment, but it is enough of an interface to start getting the important parts off the ground: an interface to Circuit that allows for dynamic disconnections, which will provide for most of the simulation logic.

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+	
+	internal fun removeVertex(v: Vertex) {
+		vertexListeners.forEach { it.vertexRemoved(v) }
+		disconnect(v)
+		vertices.remove(v)
+	}
+
+	internal fun connect(a: Vertex, b: Vertex): Edge {
+		assert(a.graph == this) { "CG.c: vertex a=$a not in graph $this" }
+		assert(b.graph == this) { "CG.c: vertex b=$b not in graph $this" }
```
jrddunbr, line 82: I think we have cleaner debug prints but perhaps it would be worth having a `dsf()` (debug string format) call to do this cleanly and consistently.

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+		b.incident[a] = edge
+
+		edgeListeners.forEach { it.edgeAdded(edge) }
+		
+		return edge
+	}
+	
+	internal fun disconnect(a: Vertex, b: Vertex): Boolean {
+		assert(a.graph == this) { "CG.d: vertex a=$a not in graph $this" }
+		assert(b.graph == this) { "Cg.d: vertex b=$a not in graph $this" }
```
jrddunbr, line 96: dsf

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+		upPhase: (visited: Vertex, depth: Int) -> Unit,
+		start: Vertex? = null
+	) {
+		val unvisited = vertices.toMutableSet()  // XXX Relying on the implementation detail of a copy here...
+		var depth = 0
+		val stack = Stack<Visitation>()
+
+		while(!unvisited.isEmpty()) {
+			if(stack.isEmpty()) {
+				dprintln("CG.vDF: adding a new root")
```
jrddunbr, line 136: Use new debug stuff

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+				stack.push(Visitation(start ?: unvisited.first(), Visitation.Phase.Down, true))
+			}
+
+			val visitation = stack.pop()
+			val vertex = visitation.vertex
+			dprintln("CG.vDF: visiting $vertex phase ${visitation.phase} root ${visitation.root}")
```
jrddunbr, line 142: ''

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+
+			when(visitation.phase) {
+				Visitation.Phase.Down -> {
+					unvisited.remove(vertex)
+					stack.push(Visitation(vertex, Visitation.Phase.Up))
+
+					val neighbors = vertex.incident.keys
+					dprintln("CG.vDF: pre-down neighbors: $neighbors")
```
jrddunbr, line 150: ''

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+					downPhase(vertex, depth, visitation.root, neighbors)
+					dprintln("CG.vDF: post-down neighbors: $neighbors")
```
jrddunbr, line 152: ''

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+	// FIXME: There are several optimizations which can be applied here:
+	// - Find bridges (non-bridge edges can be removed without issue)
+	// - Minimize component changeover (don't just choose one and clear it)
+	internal fun assignComponents() {
+		var currentComponent: ConnectedComponent? = null
+		val newComponents: MutableSet<ConnectedComponent> = mutableSetOf()
+
+		visitDepthFirst({
+			vertex, _, isRoot, _ ->
+			dprintln("CG.aC: visiting $vertex root=$isRoot")
```
jrddunbr, line 173: ''

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+				newComponents.add(currentComponent!!)
+				currentComponent!!.clear()
+				currentComponent!!.add(vertex)
+			} else {
+				vertex.component = currentComponent!!
+				currentComponent!!.add(vertex)
+			}
+		}, {_, _ -> Unit})
+
+		dprintln("CG.aC: old components $components, new components $newComponents")
```
jrddunbr, line 190: ''

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+	
+	var outstandingGuards = 0
+		protected set
+
+	inner class MutationGuard internal constructor() {
+		init {
+			outstandingGuards++
+			dprintln("CG.MG.<init>: beginning mutation, nest level $outstandingGuards")
```
jrddunbr, line 217: ''

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+		}
+		
+		fun connect(a: Vertex, b: Vertex) = this@ComponentGraph.connect(a, b)
+		fun disconnect(a: Vertex, b: Vertex) = this@ComponentGraph.disconnect(a, b)
+		fun disconnect(v: Vertex) = this@ComponentGraph.disconnect(v)
+		fun newVertex() = this@ComponentGraph.newVertex()
+		fun removeVertex(v: Vertex) = this@ComponentGraph.removeVertex(v)
+		
+		fun drop() {
+			dprintln("CG.MG.drop: ending mutation, nest level $outstandingGuards")
```
jrddunbr, line 227: ''

core/src/test/kotlin/org/eln2/data/ComponentGraphTest.kt
```diff
@@ -0,0 +1,303 @@
...
+    
+    fun fullConsistencyCheck(cg: ComponentGraph) {
+        try {
+            fullConsistencyCheckImpl(cg)
+        } catch(e: Throwable) {
+            DEBUG = true
+            System.err.println("The following graph failed consistency checks:")
+            debugDump(cg)
+            throw e
+        }
```
jrddunbr, line 69: This should be an assert plus a debug print, rather than a throw and an error print.

jrddunbr (changes requested):
I'm bad at this sort of code review because I can't quite visualize what is going on (plus it's 11pm and I'm tired), but for starters, let's look at fixing some debug statements.

core/src/test/kotlin/org/eln2/data/ComponentGraphTest.kt
Grissess: The `try/catch` here actually catches assertions and rethrows them; it's `fullConsistencyCheckImpl`'s job to raise an assertion whenever it finds an inconsistency. The problem is that the assertion error doesn't have the nice full graph, so I conditionally dump the _entire_ graph whenever a consistency check fails. Due to the reliance on the "old" debug architecture and tons of `dprintln`, I forced debugging on globally as a bit of a last-minute dirty hack to ensure that the graph _would_ make it to the log.

This infrastructure has been _extremely_ helpful for debugging issues with the data model, but it definitely still is a hack. I haven't looked at the new debug system, but I'm hoping that would fix this :)

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
Grissess: I'm looking at `master` for the "new debug stuff", and not seeing much in the way of substantial difference (mostly, it looks like it automatically extracts the calling class/method name). Is that what you're talking about?

core/src/test/kotlin/org/eln2/data/ComponentGraphTest.kt
jrddunbr: The point of using `dprintln()` is to get it into the signature. We should be using it everywhere.

The reason I want an assert is because I *don't* want a re-throw. If you need more details, print the information you want to see before calling `assert()`. This will end the test early, and show the error promptly.

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+	internal open fun clear() = vertices.clear()
+
+	internal var realized = false
+	internal open fun realize() {}
+	internal open fun unrealize() {}
+}
+
+interface ComponentListener {
+	fun componentRealized(c: ConnectedComponent)
+	fun componentUnrealized(c: ConnectedComponent)
```
jrddunbr, line 30: I'm not sure what realized is in this context, and these need comments.

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,278 @@
...
+	fun addListener(listener: Any) {
+		if(listener is ComponentListener) componentListeners.add(listener)
+		if(listener is VertexListener) vertexListeners.add(listener)
+		if(listener is EdgeListener) edgeListeners.add(listener)
+	}
+	
+	fun removeListener(listener: Any) {
+		componentListeners.remove(listener)
+		vertexListeners.remove(listener)
+		edgeListeners.remove(listener)
```
jrddunbr, line 61: Checking the type will make this faster, don't try this operation needlessly on 3 lists. That said, does the compiler optimize this? Not sure it does.

jrddunbr (changes requested):
Can you explain what the expected API interface is of this class? A lot of things are internal and open and it confuses me how you even interface with this class, since there is no documentation. Please add a bunch of Javadoc on anything public, and preferably also to things internal.

jrddunbr:
I'm updating the branch because I think master has a fix for some of the failed builds.

jrddunbr:
@Grissess status on this?

Grissess:
What's wrong with the debug prints, again? :)

build.gradle.kts
```diff
@@ -76,6 +76,8 @@ subprojects {
                 showExceptions = true
                 showCauses = true
                 showStackTraces = true
+                // TODO: add the other envvars that can trigger debug output
+                showStandardStreams = System.getenv("MODS_ELN_DEBUG") != ""
```
jrddunbr, line 80: We have more than just this. We could detect all of them by hooking into the debug code? Perhaps for another PR. Let's keep this PR just to CG.

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
```diff
@@ -0,0 +1,273 @@
+package org.eln2.data
+
+import org.eln2.debug.dprintln
+import java.lang.StringBuilder
+import java.util.*
```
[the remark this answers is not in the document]
jrddunbr: Fix plz

core/src/main/kotlin/org/eln2/data/ComponentGraph.kt
jrddunbr: https://github.com/eln2/eln2/blob/master/core/src/main/kotlin/org/eln2/debug/DebugPrint.kt

TL;DR, remove the `CG.vDF` because we do that for you now.

core/src/test/kotlin/org/eln2/data/ComponentGraphTest.kt
jrddunbr: I'll let this slide for now.

jrddunbr (changes requested):
Just remove the characters at the beginning of your printlns because we have those already handled by the println utility here: https://github.com/eln2/eln2/blob/master/core/src/main/kotlin/org/eln2/debug/DebugPrint.kt

That and we should change the `build.gradle.kts` file in a separate PR, and handle all the different strings that the debugging code picks up on.

build.gradle.kts
Grissess: I needed it for this PR, so I stuck it in there for only that reason. I can totally back it out though :P

Grissess:
As for the other two that I can't seem to comment on: I did notice that prefix to the regular ol' debug print facility, and I approve; the only thing I'm even slightly concerned about is the runtime cost of that, but that's another axe to grind. You'll notice I used abbreviations, too, to keep the column width of debug prints as small as reasonable, but... once again...

As long as that's the only thing troubling you, that's easy enough!

And, for the star import--do you have any IDEA tricks for that? I think IDEA itself just decided to star it because I was using "enough" imports from `java.util` (for some definition of "enough") that it thought it was being helpful. I probably can resolve that manually, but I'm worried IDEA will just star it again if I use its own facilities to resolve those imports :P

build.gradle.kts
jrddunbr: I guess it's fine, but flesh it out I guess?

jrddunbr:
I would highly recommend telling IDEA not to do asterisks ever. It makes the imports easier to read. The setting is somewhere in the settings menu, which is searchable.

I would change it for Kotlin and Java if you want.

jrddunbr:
RE: Debug prints taking a long time - meh. Printing by itself takes a little while as it is since it's buffered and stuff, I don't think spending a few extra cycles to find the class name is going to be too bad. Besides, it's debugging code, it shouldn't ever need to run while the game is running unless you've turned it on for some reason.

Grissess:
image

Well now, I found the culprit for Java, but I can't find it for Kotlin... here's to hoping it's the same setting for both :)

Grissess:
:tada:

PR #72, opened 2020-05-23
Add some failing tests to fix

Grissess:
This PR is poised to eventually fix some new issues discovered in unification of circuit `Node`s.


core/src/main/kotlin/org/eln2/sim/electrical/mna/component/Component.kt
```diff
@@ -116,16 +130,18 @@ abstract class Component : IDetail {
...
-		} else {
-			nr.node = nodes[nidx].node
-		}
+		val (nnamed, tnnamed) = Pair(n.nameSet, tn.nameSet)
+		if(nnamed && !tnnamed) tn.named(n.name)
+		if(tnnamed && !nnamed) n.named(tn.name)
+		tn.unite(n)
+		nodes[nidx].node = tn
+
+		// Assertion intended--fail loudly if circuit mutated here.
```
jrddunbr, line 139: I'm thinking that it would be better if we handled this more gracefully; we're talking a full halt of an entire server over one bad circuit here.

jrddunbr (changes requested):
This is just one circuit in a sea of thousands, if one circuit goes bad, we should inform the server in a more recoverable way than just throwing a more generic Kotlin type error.

core/src/test/kotlin/org/eln2/sim/electrical/mna/CircuitTest.kt
```diff
@@ -70,6 +93,86 @@ internal class CircuitTest {
...
+		ts.c.step(0.05)
+		assertEquals(ts.c1.current, 0.000732, 0.000001)
+		assertEquals(ts.c1.potential, 6.315, 0.001)
+		ts.c.step(0.05)
+		assertEquals(ts.c1.current, 0.000270, 0.000001)
+		assertEquals(ts.c1.potential, 8.660, 0.001)
+		ts.c.step(0.05)
+		assertEquals(ts.c1.current, 0.000099, 0.000001)
+		assertEquals(ts.c1.potential, 9.506, 0.001)
+	}
```
jrddunbr, line 175: @Grissess these tests are backwards... it is `expected`, `actual`, `interval`.

Grissess: That argument order annoys me--I'd much rather rebind the function swapping arguments than rewrite those tests. (That's how you read equalities like this: "assert [value under test] equals [compared value"--unless you happen to be one of those "defensive" programmers that deliberately avoids putting an lvalue on the left hand side so as to accidentally catch "foo = 7" assignment expression bugs.)

The function doesn't change in either case :)

core/src/main/kotlin/org/eln2/sim/electrical/mna/component/Component.kt
Grissess: It's not really a "bad circuit" as "totally illegal program state". This should never happen, and if it does, the error is intentional.

Making it catchable is a good compromise.

core/src/test/kotlin/org/eln2/sim/electrical/mna/CircuitTest.kt
jrddunbr: No, the function doesn't change, but the debug printouts do. If the order doesn't functionally change anything, I'd much rather have the proper assertions using the code as designed, not as convenient, that way it debug prints properly.

core/src/main/kotlin/org/eln2/sim/electrical/mna/component/Component.kt
jrddunbr: Make it so.

core/src/test/kotlin/org/eln2/sim/electrical/mna/CircuitTest.kt
jrddunbr: Which reminds me - in my tests I handled it differently - I handled everything as a boolean statement, and then I just did my own statement in there and made sure that if things were good that it passed the assertion. This had the added benefit of allowing me to debug this in the testing logic, rather than dying in the test assertions and not being able to output debugging information. Arguably here, the best solution would be to change how those tests are run so that if it plans to fail, that it instead exports the graph of what it was doing and *then* it calls in the assertion and fails. That allows us to have better control over our debug logic and include debugging output as part of the testing infrastructure.

Caeleron:
After comparing what we have to what Falstad has, it seems that the variables that handle the current across a capacitor and how much current the associated current source draw is the same in ours but different in Falstad's. This is the only difference that I am seeing between the implementations, but there's works while ours somehow doesn't. The behavior I'm currently seeing with capacitors is that they at first report "-0.0" current over them for the first step they are added, and then the same value for all subsequent steps. Voltage sources attached to capacitors report the correct current on the first step, but then don't change at all. I am unsure of how to fix this since there doesn't appear to be any errors on the surface, but its failing every test using a capacitor for this reason.

core/src/test/kotlin/org/eln2/sim/electrical/mna/CircuitTest.kt
jrddunbr: Cael suggested this may be a bit over-the-top, so changing the tests to use the methods correctly would be sufficient I think.

Grissess:
If that looks like the implementation really is in parity, I'm going to suspect the infrastructure is to blame--contributions to the matrices might not be made that ought to be. This warrants some further investigation.

jrddunbr:
@Grissess Cael said he expects you will investigate why this is broken and intends to leave it to you.

My cursory inspection shows that your capacitor code does not implement something related to the current source correctly; I think it should be separate from  the overall current across the device.

jrddunbr:
It is apparent to me that the MNA is pretty broken. At this point, the only thing that we've done is improve the code and mark with a test that our capacitor implementation is (has been) broken.

jrddunbr (approved):

PR #73, opened 2020-05-25
Adds ktlinter via reviewdog

jrddunbr:
A new Kotlin linter to try out.

### DIFF
--- a/.github/workflows/ktlint.yml
+++ b/.github/workflows/ktlint.yml
@@ -1,0 +1,18 @@
+name: ktlint
+on: [pull_request]
+jobs:
+  ktlint:
+    name: Check Code Quality
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jrddunbr:
So I feel that Codacy is not doing a very great job. I'd like to replace it.

I know, we have a lot of our tests such as coverage and error checking in there, but if I'm quite honest, the only thing Codacy really does well is code coverage.

The thing I like about this tool is that it can do review comments on PR's directly so that you know exactly what issues there are without going into the Codacy UI to find what files are causing the issue - this is a problem because every time I try to drill down on the issue, their UI is either slow or cumbersome.

I'm not certain I configured it right, but I did try to get it into at least a usable state.

Baughn:
Well, let's try it.

PR #74, opened 2020-05-25
Ktlintfix

Baughn:
Fixes most of the lints introduced by ktlint.

This forces us to use spaces, not tabs. Gah.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,8 +14,14 @@ root = true
 end_of_line = lf
 insert_final_newline = true
 charset = utf-8
+max_line_length = 80
 
 # Use a common indentation style for all our apps.
-# It's tab because anyone weak-sighted needs the ability to override indentation.
-indent_style = tab
+indent_style = space
 indent_size = 4
+
+[*.{kt,kts}]
+max_line_length = 120
+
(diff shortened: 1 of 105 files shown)


jrddunbr:
GitHub dumped the entire rest of the output of this linter into my email. And then Gmail told me the mail was too large for my inbox and that I had to read the raw text.

Proposal:
* Disable 120 character limit
* Disable "Missing spacing before < and > and { and  } and :"
* Disable "Imports must be ordered in lexicographic order without any empty lines in-between"

You already know about the tabs issue.

jrddunbr:
Perhaps changing from `info` to `warning` in the linter would also be a good idea. Not sure if that would catch documentation errors though (or if it does right now, which is honestly most of what it does anyways)

Baughn:
The idea is, I believe, that the linter should only catch lint errors introduced by any given PR.

This one touches *everything*, so it falsely thinks it introduces all the errors, but future PRs won't keep showing the same lint messages.

jrddunbr:
Ahhh, that makes sense.

jrddunbr:
Here goes nothing!

PR #77, opened 2020-05-28
Fixing gradle and modbus

rbb30:
Made gradle work with kotlinforforge 1.2.1 and changed FMLJavaModLoadingContext to kotlinforforge's MOD_CONTEXT as recommended.

PR #81, opened 2020-06-01
Include KFF and fix Minecraft standalone builds dependency issues

### LINKED ISSUE #76: Eln2 Minecraft Integration Jar not shadowed correctly
If I build the Eln2 Minecraft Integration jar, it does not include the necessary Kotlin and other resources.

jrddunbr:
Scope of PR:

- [x] Include KFF as a build dep
- [x] Set dependencies for Minecraft on KFF and Bookshelf
- [ ] Create shadowed Minecraft jar that can be placed into a Minecraft launcher manually with the two library mods and launched without crashes
- [ ] (optional) Allow uploading to CurseForge from the build env (publish task)

Per @Baughn: We may not actually want to use KFF as a build dep, but still have it as a minecraft mod dep so that it pulls in Kotlin runtimes into our classpath for us to use.


integration-mc1-15/src/main/kotlin/org/eln2/Eln2.kt
```diff
@@ -28,9 +27,8 @@ object Eln2 {
         // Register blocks and items
         ModItems.values().forEach { registry.registerItem(it.items, it.name.toLowerCase()) }
         ModBlocks.values().forEach { registry.registerBlock(it.block, it.name.toLowerCase()) }
-
-        registry.initialize(FMLJavaModLoadingContext.get().modEventBus)
-    }
+		registry.initialize(thedarkcolour.kotlinforforge.forge.MOD_BUS);
```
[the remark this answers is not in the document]
jrddunbr: fixed

jrddunbr:
I closed #77 and opened this because #77 was no longer mergable and I wanted to bring the code back into this from the forked branch.

There is still work to be done, but there are no longer critical merge conflicts.

jrddunbr:
One of the tests is failing in master, so the CI failing is expected.

integration-mc1-15/src/main/java/org/eln2/Eln2Wrapper.java
```diff
@@ -22,29 +21,29 @@
...
+    public Eln2Wrapper() {
+        KotlinEventBus bus = thedarkcolour.kotlinforforge.forge.ForgeKt.getMOD_BUS();
+        // Register the setup method for modloading
+        bus.addListener(this::loadComplete);
+        // Register the enqueueIMC method for modloading
+        bus.addListener(this::enqueueIMC);
+        // Register the processIMC method for modloading
+        bus.addListener(this::processIMC);
+        // Register the doClientStuff method for modloading
+        bus.addListener(this::doClientStuff);
```
jrddunbr: Do these comments really add any value?

integration-mc1-15/src/main/kotlin/org/eln2/Eln2.kt
```diff
@@ -25,17 +30,30 @@ object Eln2 {
...
+        ModBlocks.values().forEach {
+            registry.registerBlock(it.block, it.name.toLowerCase())
+        }
+    }
+
+    fun itemRegistry(event: RegistryEvent.Register<Item>){
+        ModItems.values().forEach {
+            registry.registerItem(it.items, it.name.toLowerCase())
+        }
+    }
```
jrddunbr: Is there any benefit to splitting this out into functions? I don't think there is.

Also, I want to avoid using KFF, because I'm on the fence of even including it, rather than shadowing in Kotlin ourselves. Because it doesn't even show up as a mod dependency in Minecraft, I feel that it is non-user intuitive and not a good thing for us to use. We're already going to have to shadow our application jars, and the textures and stuff are going to make the jars pretty large anyways.

jrddunbr: Hmm. If these functions are called by annotations, it's very unclear that is happening to me.

jrddunbr:
According to Codacy, `import net.minecraftforge.fml.event.lifecycle.*;` in Eln2.java is unused.

jrddunbr:
@lolmegaxd1 so here's an interesting thing. Look at the branch state/exportworks, you'll notice that it both runs in the dev environment, as well as produces a working jar.

Worth noting, I did comment the KFF dependency line in the mods.toml, but I'm not certain that change is actually what did it - I think that changing that one shadow line fixes the code and that the KFF change was actually unnecessary.

On a separate note, it's possible that the KFF dependency is broken anyways, because Forge does not consider it to be a mod for some reason as far as dependency resolution is concerned. If that is actually the case, I'd much rather we get rid of KFF as a dependency at all, and just shadow Kotlin ourselves.

jrddunbr:
I hate to say it but this is taking long enough that it becomes un-rebasable. I think at the end this one will need to be merge committed rather than rebased due to the number of commits on there.

jrddunbr:
We fixed the issue with the testing infrastructure so this should compile and stuff now maybe? This still needs to be tested externally of the testing system on someone's desktop because we don't do integration tests for jar generation.

jrddunbr:
This code is for 1.15.2 and we're moving to 1.16.1 or newer. I'm going to be refactoring the code as a result to support 1.16.1 and that's going to make this stuff not mergable. I'm going to close this and fix it as part of that new migration.

PR #82, opened 2020-06-01
Fixes lints in files related to sim code

jrddunbr:
There were a handful of lints that I fixed through much of the code.

It makes it much easier to understand `Circuit` as well if we have the internal stuff marked `internal`.

I also corrected some nested `it` situations, which often lead to code errors.

### DIFF
--- a/core/src/main/kotlin/org/eln2/sim/electrical/mna/component/Component.kt
+++ b/core/src/main/kotlin/org/eln2/sim/electrical/mna/component/Component.kt
@@ -90,6 +90,7 @@ abstract class Component : IDetail {
      * Component implementations should index this list and avoid storing references anywhere else. The owning
     Circuit holds all Nodes it grants weakly, so taking other references will result in a memory leak.
      */
+    @Suppress("LeakingThis") // Safe in a single threaded setting, apparently.
     var nodes: MutableList<NodeRef> = ArrayList(nodeCount)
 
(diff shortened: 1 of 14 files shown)


core/src/main/kotlin/org/eln2/sim/electrical/mna/component/Component.kt
```diff
@@ -90,19 +90,21 @@ abstract class Component : IDetail {
      * Component implementations should index this list and avoid storing references anywhere else. The owning
     Circuit holds all Nodes it grants weakly, so taking other references will result in a memory leak.
      */
+    @Suppress("LeakingThis") // Safe in a single threaded setting, apparently.
```
jrddunbr, line 93: @Baughn I'm curious what this lint was all about. On Stack Overflow, they say that in cases where the constructor is not multi-threaded, there is no issue. I didn't read into it too much since `Circuit` (and thus, `Component`) are not threaded, but this may cause other issues. This sets the size of the ArrayList of nodes and voltage sources, but I think that if someone tries to access them by index, they may be too small and cause errors such as out of bounds array index. I'm not certain how to proceed, but I think this may need a refactor or a pre-medetated large value.

jrddunbr:
CI test failure to be expected since RC circuit test is still failing. Nothing else failed in the tests.

PR #83, opened 2020-06-01
MNA Hotfixes, Round 2

jrddunbr:
This is all @Grissess fault :P

These are all supposed to be part of the "recently" merged mna hotfix branch.


Grissess:
For the record, this patch on the hotfix:

- Removes `NodeRef` (it's not needed anymore now that `Node`s are `Set`s);
- Adds `priority` to `Set` (to implement the `GroundNode` merge proper);
- Updates the documentation.

As a bit of an afterthought, now that this is merged in master, `Set` gets moved to `data` where it belongs.

PR #87, opened 2020-06-03
Adds MatrixPrint and some tests for MNA

jrddunbr:
I added some new debug printing/formatting tools for dumping the
matricies from the circuit into the console in a readable format.

I also added a set of resistor tests which appears to show a bug with
the copying of data from the MNA unknowns matricies into the correct
nodes.


core/src/main/kotlin/org/eln2/sim/electrical/mna/Circuit.kt
```diff
@@ -177,7 +177,7 @@ class Circuit {
      *
      * It is the "A Matrix" in the [MNA Algorithm][https://lpsa.swarthmore.edu/Systems/Electrical/mna/MNA3.html].
      */
-    internal var matrix: RealMatrix? = null
+    var matrix: RealMatrix? = null
```
jrddunbr, line 180: Change back to internal

core/src/main/kotlin/org/eln2/sim/electrical/mna/Circuit.kt
```diff
@@ -186,7 +186,7 @@ class Circuit {
      *
      * This is the "z matrix" in the [MNA Algorithm][https://lpsa.swarthmore.edu/Systems/Electrical/mna/MNA3.html].
      */
-    internal var knowns: RealVector? = null
+    var knowns: RealVector? = null
```
jrddunbr, line 189: Change back to internal

core/src/main/kotlin/org/eln2/sim/electrical/mna/Circuit.kt
```diff
@@ -468,11 +468,11 @@ class Circuit {
      * This routine can fail; [success] is set based on whether this method was successful.
      */
     // Step 3: With known current and voltage sources, solve for unknowns (node potentials and source currents).
-    private fun computeResult() {
+    internal fun computeResult(): RealVector? {
```
jrddunbr, line 471: I added a return here so that you can dump the `unknowns` matrix out of this and into a testing class.

PR #93, opened 2020-06-05
Adds POSITIVE and NEGATIVE as pin aliases

jrddunbr:
Also adds new tests for capacitor and inductor that fail :D


Grissess:
Psst: POSITIVE and NEGATIVE want to be in `Component.kt`, near the definition of `Pole`, which defines the `pos` and `neg` fields :)

jrddunbr:
Make it so?

PR #94, opened 2020-06-05
Create main.kt

gugus:
very basic thermal simulation

### DIFF
--- a/core/src/main/kotlin/org/eln2/sim/thermal/main.kt
+++ b/core/src/main/kotlin/org/eln2/sim/thermal/main.kt
@@ -6,11 +6,11 @@ import kotlin.math.*
 /**
  * represents a thermal interface between two objects. At each tick, the flow can be marked as computed.
  */
-class ThermalInterface{
+class ThermalInterface {
 
     var mutex = Mutex()
-    var R = 1f //heat resistance at the interface
-    var heatTranfer = 0f //heat transfer during a simulation tick through this interface
+    var R = 1.0 //heat resistance at the interface
+    var heatTranfer = 0.0 //heat transfer during a simulation tick through this interface
 
     var lastTickComputed = 0 //number of the last tick where the heat transfer was computed. Tick number switch between 0 and 1. This is here to prevent computing twice the heat trasnfered when not needed
 
@@ -18,7 +18,7 @@ class ThermalInterface{
      * compute the heat transfer from object 0 to object 1 during time dt for a given tick.
      * If the flow was already computed, use the stored value.
      */
-    fun computeHeatTransfer(dt : Float, tickNumber : Int){
+    fun computeHeatTransfer(dt: Double, tickNumber: Int) {
         var T0 = e0.getT(tickNumber)
         var T1 = e1.getT(tickNumber)
         if(tickNumber != lastTickComputed) {
@@ -27,24 +27,24 @@ class ThermalInterface{
         }
     }
 
-    lateinit var e0 : ThermalElement
-    lateinit var e1 : ThermalElement
+    lateinit var e0: ThermalElement
+    lateinit var e1: ThermalElement
 }
 
 /**
  * a thermal interface computes the flow from object 0 to 1. Adding a sign to it allow us
  * to represent the flow from 1 to 0 without having to compute it twice
  */
-class SignedThermalInterface(thInterface : ThermalInterface, sgn : Float){
-    var mThermalInterface :ThermalInterface = thInterface
-    var sign : Float = sgn
+class SignedThermalInterface(thInterface: ThermalInterface, sgn: Double) {
+    var mThermalInterface: ThermalInterface = thInterface
+    val sign: Double = sgn
 
     /**
      * compute the heat transfer during time dt for a given tick
      */
-    fun computeHeatTransfer(dt : Float, tickNumber : Int):Float{
-        mThermalInterface.computeHeatTransfer(dt,tickNumber)
-        return sign*mThermalInterface.heatTranfer;
+    fun computeHeatTransfer (dt : Double, tickNumber : Int):Double{
+        mThermalInterface.computeHeatTransfer(dt, tickNumber)
+        return sign * mThermalInterface.heatTranfer;
     }
 }
 
@@ -52,21 +52,21 @@ class SignedThermalInterface(thInterface : ThermalInterface, sgn : Float){
  * a ThermalElement represents a physical object with a given temperature and thermal capacity.
  * It also posses a list of signed thermal interfaces.
  */
-class ThermalElement{
-    var C = 1f; //Heat capacity of a given element
-    var P = 0f; //Power dissipated as heat in the thermal element (from electrical circuit)
+class ThermalElement {
+    var C = 1.0 //Heat capacity of a given element
+    var P = 0.0 //Power dissipated as heat in the thermal element (from electrical circuit)
 
     //temperature needs to be stored twice: at time $t$, we use T0, then we use that to
     //compute temperature at time $t+dt$ that we store in T1, then for $t + dt+dt$ we
     //store it in T0 again etc
-    var T0 = 300f //in Kelvin
-    var T1 = 300f
+    var T0 = 300.0 //in Kelvin
+    var T1 = 300.0
     //var useT0T1 = 0
 
     /**
      * returns the temperature for the current tick
      */
-    fun getT(tickNumber: Int):Float{
+    fun getT(tickNumber: Int):Double {
         if(tickNumber.rem(2) == 0){
             return T0
         }
@@ -78,18 +78,18 @@ class ThermalElement{
     /**
      * Update the temperature for the current tick for an update in time of dt.
      */
-    suspend fun updateT(dt : Float, tickNumber: Int){
+    suspend fun updateT(dt: Double, tickNumber: Int) {
         //println("c")
-            for (signedInterface in themalInterfacesList) {
-                signedInterface.mThermalInterface.mutex.withLock {
-                    var J = signedInterface.computeHeatTransfer(dt, tickNumber)
-                    if (tickNumber.rem(2) == 0) {
-                        T1 = T1 + J / C + P * dt
-                    } else {
-                        T0 = T0 + J / C + P * dt
-                    }
+        for (signedInterface in themalInterfacesList) {
+            signedInterface.mThermalInterface.mutex.withLock {
+                var J = signedInterface.computeHeatTransfer(dt, tickNumber)
+                if (tickNumber.rem(2) == 0) {
+                    T1 = T1 + J / C + P * dt
+                } else {
+                    T0 = T0 + J / C + P * dt
                 }
             }
+        }
     }
 
     var themalInterfacesList = ArrayList<SignedThermalInterface>()
@@ -99,20 +99,20 @@ class ThermalElement{
 /**
  * helper class that creates thermal elements / interfaces from real world physical values
  */
-class ThermalBuilder{
+class ThermalBuilder {
 
     /**
      * creates a thermal element corresponding to a copper Wire
      */
-    fun createWireThermalElement(length: Float, radius : Float):ThermalElement{
+    fun createWireThermalElement(length: Double, radius: Double):ThermalElement {
         var wire = ThermalElement()
 
         var volume = length * PI * radius*radius
-        var volumicMass = 8940f //copper volumic mass in kg/m^3
+        var volumicMass = 8940.0 //copper volumic mass in kg/m^3
         var mass = volume * volumicMass
-        var heatCapacityPerKilogram = 385f //copper heat capacity per Kg in J/Kg/K
+        var heatCapacityPerKilogram = 385.0 //copper heat capacity per Kg in J/Kg/K
 
-        wire.C = (mass*heatCapacityPerKilogram).toFloat()
+        wire.C = mass*heatCapacityPerKilogram
         println("C")
         println(wire.C)
         return wire
@@ -123,12 +123,12 @@ class ThermalBuilder{
 /**
  * main thermal simulation class
  */
-class ThermalSimulator{
+class ThermalSimulator {
 
     /**
      * update the thermal simulation (i.e. every thermal element). Should be parellel with coroutines.
      */
-    suspend fun updateSimulation(dt : Float,tickNumber : Int) = runBlocking{
+    suspend fun updateSimulation(dt: Double,tickNumber: Int) = runBlocking{
         for(l in L){
             //println("a")
             val c = launch {
@@ -152,9 +152,9 @@ fun main() = runBlocking<Unit> {
     var builder = ThermalBuilder()
     //sim.L = L
     for(i in 0..100){
-        var a = builder.createWireThermalElement(1f,0.01f)//ThermalElement()
-        a.T0 = 300f + 10f * i
-        a.T1 = 300f + 10f * i
+        var a = builder.createWireThermalElement(1.0,0.01)//ThermalElement()
+        a.T0 = 300.0 + 10.0 * i
+        a.T1 = 300.0 + 10.0 * i
         sim.L.add(a)
 
         //add cpnnection with previous element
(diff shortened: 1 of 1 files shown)


core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
+import kotlinx.coroutines.*
+import kotlinx.coroutines.sync.*
+import kotlin.math.*
+
+class ThermalInterface{
```
Grissess, line 5: Little things: it looks better to leave a space around most punctuation, including braces, separating it from alphanumerics.

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
...
+
+    var mutex = Mutex()
+    var R = 1f //heat resistance at the interface
+    var heatTranfer = 0f //heat transfer during a simulation tick through this interface
+
+    var lastTickComputed = 0 //number of the last tick where the heat transfer was computed. Tick number switch between 0 and 1. This is here to prevent computing twice the heat trasnfered when not needed
```
Grissess, line 11: If the domain is finite _and_ size two, do you think a boolean would work better here?

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
...
+            heatTranfer = (T1 - T0) / R * dt
+            lastTickComputed = tickNumber
+        }
+    }
+
+    lateinit var e0 : ThermalElement
+    lateinit var e1 : ThermalElement
+}
+
+class SignedThermalInterface(thInterface : ThermalInterface, sgn : Float){
```
Grissess, line 27: You can integrate these fields right into the constructor parameters if you don't need to do anything fancy with them:

```kotlin
class SignedThermalInterface(val thermalInterface: ThermalInterface, val sign: Float) {
    fun computeHeatTransfer(dt: Float, tickNumber: Int): Float {
        // ...
    }
}
```

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
...
+    var mThermalInterface :ThermalInterface = thInterface
```
Grissess, line 28: I notice the spacing for your colons is a bit inconsistent; I have no strong preference on which way it goes (I _think_ the style guide recommends setting it against the name, with a space between it and the type), consistency will really help my poor eyes :)

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
...
+    var sign : Float = sgn
```
Grissess, line 29: ...this only needs to be a `val`, right?

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
...
+
+    fun computeHeatTransfer(dt : Float, tickNumber : Int):Float{
```
Grissess, line 31: One thing I noticed quite a lot through this section: there's a lot of `Float` here. I dislike that even the JVM makes you pick between Float and Double (it doesn't matter to Python, for example), but the space concerns are handled at the conversion boundaries, and there's no real performance benefit to using a float over a double in a managed language such as the JVM. For consistency's sake, our matrix math library seems to use doubles mostly, so...

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
...
+        mThermalInterface.computeHeatTransfer(dt,tickNumber)
+        return sign*mThermalInterface.heatTranfer;
```
Grissess, line 33: Again, little nits: that comma could use a space after it, and that asterisk a space on either side :)

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+class ThermalElement{
+    var C = 1f; //Heat capacity of a given element
+    var P = 0f; //Power dissipated as heat in the thermal element (from electrical circuit)
```
Grissess, line 39: You shouldn't need the semicolons here; the Kotlin style guide seems to advise against them unless necessary

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
...
+
+    //temperature needs to be stored twice: at time $t$, we use T0, then we use that to
+    //compute temperature at time $t+dt$ that we store in T1, then for $t + dt+dt$ we
+    //store it in T0 again etc
+    var T0 = 300f
+    var T1 = 300f
+    //var useT0T1 = 0
+
+    fun getT(tickNumber: Int):Float{
+        if(tickNumber.rem(2) == 0){
```
Grissess, line 49: Correct me if I'm wrong: is that just the `%` operator?

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,146 @@
...
+//helper class that creates thermal elements / interfaces from real world physical values
+class ThermalBuilder{
+
+    fun createWireThermalElement(length: Float, radius : Float):ThermalElement{
+        var wire = ThermalElement()
+
+        var volume = length * PI * radius*radius
+        var volumicMass = 8940f //copper volumic mass in kg/m^3
+        var mass = volume * volumicMass
+        var heatCapacityPerKilogram = 385f //copper heat capacity per Kg in J/Kg/K
```
Grissess, line 84: I foresee someone eventually wanting to abstract away these two parameters, but thank you for finding them :)

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
gugus: It should work the same, but I find it easier to read that the tickNumber is different. It might also help with debugging since you can actually keep track of the ticknumber if you want to, instead of being restricted to 0 and 1. 

That being said, I don't really have a strong opinion on that

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
gugus: I had no idea you could do that! Is that the common practice?

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
gugus: Noted :)

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
gugus: You are right

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
gugus: Okay I'll switch to double. I was wondering about that too.

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
gugus: done

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
gugus: C++ habits? :p

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
gugus: Maybe? I searched on the internet for modulo in Kotlin and I got that operator. Maybe % also work but I didn't find any reference about it.

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
gugus: Yup, I'm planning to write some kind of map where you can look for these values for any material we use. For now, I hard coded copper just to show how the API could look like

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
Grissess: C, C++, Rust... I do the same thing when I'm switching between languages too :P

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
Grissess: https://kotlinlang.org/docs/reference/operator-overloading.html#arithmetic seems to imply the operator and method call are equivalent--overloading the method gives you the function of the operator. Thus, it should be fine :)

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
Grissess: I can't speak to how common it is in the wild, just acquainting myself with Kotlin for this project, but [it](https://github.com/eln2/eln2/blob/8a0b670a7c6bc6018e9db09a1a94aad6c32566c8/core/src/main/kotlin/org/eln2/sim/electrical/mna/Node.kt#L12) [certainly](https://github.com/eln2/eln2/blob/8a0b670a7c6bc6018e9db09a1a94aad6c32566c8/core/src/main/kotlin/org/eln2/data/MultiMap.kt#L6) [seems](https://github.com/eln2/eln2/blob/8a0b670a7c6bc6018e9db09a1a94aad6c32566c8/core/src/main/kotlin/org/eln2/space/Location.kt#L8) [useful](https://github.com/eln2/eln2/blob/8a0b670a7c6bc6018e9db09a1a94aad6c32566c8/core/src/main/kotlin/org/eln2/space/Location.kt#L41) [when](https://github.com/eln2/eln2/blob/8a0b670a7c6bc6018e9db09a1a94aad6c32566c8/core/src/main/kotlin/org/eln2/space/Location.kt#L309) [making](https://github.com/eln2/eln2/blob/8a0b670a7c6bc6018e9db09a1a94aad6c32566c8/core/src/main/kotlin/org/eln2/space/Graph.kt#L9) [containers](https://github.com/eln2/eln2/blob/8a0b670a7c6bc6018e9db09a1a94aad6c32566c8/core/src/main/kotlin/org/eln2/space/Graph.kt#L29) [or](https://github.com/eln2/eln2/blob/8a0b670a7c6bc6018e9db09a1a94aad6c32566c8/core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt#L38) [wrappers](https://github.com/eln2/eln2/blob/8a0b670a7c6bc6018e9db09a1a94aad6c32566c8/core/src/main/kotlin/org/eln2/parsers/falstad/Falstad.kt#L50). Definitely a good thing to know about :)

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
Grissess: I guess I don't either, beyond my "strongest typing" smell from other languages. Besides, managed language again; there's probably not a _whole_ lot of an optimization to be had here (if this were C, I'd definitely have dropped it into binary and/xor already :P). Whatever's clear seems good :)

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * creates a thermal element corresponding to a copper Wire
+     */
+    fun createWireThermalElement(length: Double, radius: Double):ThermalElement {
+        var wire = ThermalElement()
+
+        var volume = length * PI * radius*radius
+        var volumicMass = 8940.0 //copper volumic mass in kg/m^3
+        var mass = volume * volumicMass
+        var heatCapacityPerKilogram = 385.0 //copper heat capacity per Kg in J/Kg/K
```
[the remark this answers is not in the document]
gugus: That might require some changes to Materials.kt, am I free to make those changes?

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    var L = ArrayList<ThermalElement>()
+}
+
+/**
+ * test function
+ */
+fun main() = runBlocking<Unit> {
```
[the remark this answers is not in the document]
gugus: I'll admit, I'm 100% overwhelmed by the build system & how to integrate is properly to the project. Would it be possible for you to create an empty test function that is integrated to the build system that I can fill?

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,181 @@
...
+            //println("a")
+            val c = launch {
+                l.updateT(dt,tickNumber)
+            }
+            //println("b")
+        }
+    }
+
+    var L = ArrayList<ThermalElement>()
+}
```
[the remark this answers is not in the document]
gugus: I'm not sure what you mean here, do you mean that the updateT(dt,tickNumber) should be an IProcess?

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * represents a thermal interface between two objects. At each tick, the flow can be marked as computed.
+ */
+class ThermalInterface {
+
+    var mutex = Mutex()
+    var R = 1.0 // heat resistance at the interface
+    var heatTranfer = 0.0 // heat transfer during a simulation tick through this interface
+
+    var lastTickComputed = 0 // number of the last tick where the heat transfer was computed. Tick number switch between 0 and 1. This is here to prevent computing twice the heat trasnfered when not needed
```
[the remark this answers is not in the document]
gugus: I'll think of something for the lastTickComputed variable since IProcess only have a time delta.

Should I make the rest private or whatever Kotlin equivalent is?

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
```diff
@@ -0,0 +1,181 @@
+import kotlin.math.*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.sync.*
```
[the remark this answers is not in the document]
gugus: Will do

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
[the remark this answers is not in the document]
jrddunbr: Yes.

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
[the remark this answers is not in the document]
jrddunbr: It's very simple. Look at one of the files there ([example](https://github.com/eln2/eln2/blob/master/core/src/test/kotlin/org/eln2/compute/asmComputer/AsmComputerTest.kt)), and you will get the idea pretty quickly. Just remember to use the assertion functions to specify what exactly the results should be from the computations.

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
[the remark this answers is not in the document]
jrddunbr: At a high level, the thermal sim should be able to be called from an [IProcess](https://github.com/eln2/eln2/blob/master/core/src/main/kotlin/org/eln2/sim/IProcess.kt), yes. Your implementation of that interface will allow the MNA to basically hand of to your code and say "Hey, 30ms of time has passed, do the thermal simulation for everything you know about in that time period". Ideally, you thread it off from there inside your code, and rendezvous your internal stuff before that function exits.

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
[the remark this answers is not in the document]
jrddunbr: `private` will work until you try to access it with tests, if you need that, use `internal`.

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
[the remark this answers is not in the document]
jrddunbr: :100:

core/src/main/kotlin/org/eln2/sim/thermal/main.kt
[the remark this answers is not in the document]
gugus: Okay so the updateSimulation function should be an IProcess then right?

jrddunbr:
Moving this to #101 so that it's internal.

PR #99, opened 2020-06-30
Finishing Current Sources, Capacitors and Inductors

Caeleron:
Tests created for the following circuits:

- [x] Current Source + Resistor
- [x] RC Circuit
- [x] LR Circuit
- ~~LRC Circuit~~

If they are passing:

- [x] Current Source + Resistor
- [x] RC Circuit
- [x] LR Circuit
- ~~LRC Circuit~~ 

Components working and finished:

- [x] Current Source
- [x] Capacitor
- [x] Inductor


Caeleron:
After beginning to use real data, it appears that my test found a sweet spot in which this implementation is fairly accurate. Smaller capacitances cause the capacitor to source >1000 amps for our time step of 50 ms. Shortening the time step causes the capacitor to source too little.

Caeleron:
Meant to type "portion". Now I actually want to add a Discharging Splash Potion, since this is a MC mod ultimately.

core/src/main/kotlin/org/eln2/parsers/falstad/components/passive/CapacitorConstructor.kt
```diff
@@ -14,8 +14,8 @@ class CapacitorConstructor: PoleConstructor() {
 	override fun component(ccd: CCData) = Capacitor()
 	override fun configure(ccd: CCData, cmp: Component) {
 		val c = (cmp as Capacitor)
-		c.ts = ccd.falstad.nominalTimestep
+		c.timeStep = ccd.falstad.nominalTimestep
 		c.capacitance = ccd.data[0].toDouble()
-		c.idealU = ccd.data[1].toDouble()
+		//c.idealU = ccd.data[1].toDouble()
```
jrddunbr, line 19: Might as well delete unused code

jrddunbr (approved):
LGTM but fix this one commented line (no point in commented code)

PR #101, opened 2020-07-02
Thermal System

jrddunbr:
Pulled into our codebase. See original PR at #94 

Integration tests should be more likely to pass now.


jrddunbr:
This doesn't compile on my system, so that's going to get fixed first.

jrddunbr:
Still fails to build after I've updated it against what's in `master`.

jrddunbr:
I'm doing away with the co-routines in this code because Kotlin's implementation of co-routines is not great. I'll thread it instead using standard Java style methodology.

jrddunbr:
Yeah, we're doing the code from the ground up with some new graph theory.

core/src/main/kotlin/org/eln2/sim/thermal/ThermalNode.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import org.eln2.sim.Material
+
+class ThermalNode(
+    var material: Material,
+    var volume: Double,
+    var temperature: Double = 20.0,
+    // TODO: This is for having different surface areas depending on direction of connectivity.
+    // TODO: Not sure how we want to do that or if its already been implemented, but this is my
+    // TODO: guess on what it will look along the lines of.
```
jrddunbr, line 11: can replace with
```kotlin
/*
TODO: message here (only one TODO needed, otherwise it breaks finding one TODO per issue)

*/
```

core/src/main/kotlin/org/eln2/sim/Material.kt
```diff
@@ -61,6 +83,8 @@ Air:        0.024
 Source: http://hyperphysics.phy-astr.gsu.edu/hbase/Tables/thrcn.html
 */
 
+    abstract val specficHeat: Double
```
jrddunbr, line 86: document what the units are and why this is important to a material

core/src/main/kotlin/org/eln2/sim/thermal/ThermalEdge.kt
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+    @Synchronized
+    fun queueMoveEnergy(dt: Double) {
+        val n1 = thermalNodes.first
+        val n2 = thermalNodes.second
+
+        val x1 = conductionDistance / n1.material.thermalConductivity / conductionArea
+        val x2 = conductionDistance / n2.material.thermalConductivity / conductionArea
+        val x3 = 1.0 / 0.0005 / conductionArea
```
jrddunbr, line 36: reduce run-time math by making this one division

core/src/main/kotlin/org/eln2/sim/thermal/ThermalNode.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    val mass = volume * material.density
+
+    var energy
+        get() = mass * material.specficHeat * temperature
+        set(value) { temperature = energy / mass / material.specficHeat }
+
+    @Synchronized
+    fun commitThermalChanges() {
+        energyChanges.forEach { energy += it }
```
jrddunbr, line 25: This may not actually lock that data structure? May need to check how synchronized works

core/src/main/kotlin/org/eln2/sim/thermal/ThermalSimulator.kt
```diff
@@ -0,0 +1,63 @@
...
+            t.start()
+            threadList.add(t)
+        }
+        threadList.forEach { it.join() }
+        thermalEdges.forEach {
+            it.clearRunBit()
+        }
+    }
+
+    fun connectNodes(n1: ThermalNode, n2: ThermalNode, dir: Int, distance: Double) {
```
jrddunbr, line 36: What is the purpose of `dir`? Perhaps write some javadoc since this looks like an API

jrddunbr (commented):
Stuff and things

jrddunbr:
I'm just going to get this ready to merge because otherwise it will become unmergable.

PR #106, opened 2020-09-21
Adds some switch code.

jrddunbr:


jrddunbr:
Yo @Grissess , want to figure out how to get that SPDT switch working? We'll need components with 3 legs sooner or later.

PR #114, opened 2021-02-18
Adds interface for sim to Node

jrddunbr:
Adds an interface for the simulators into Node.

### DIFF
--- a/integration-mc1-16/build.gradle
+++ b/integration-mc1-16/build.gradle
@@ -100,6 +100,7 @@ dependencies {
     // The userdev artifact is a special name and will get all sorts of transformations applied to it.
     minecraft 'net.minecraftforge:forge:1.16.5-36.0.22'
     implementation 'thedarkcolour:kotlinforforge:1.9.0'
+    implementation project(':core')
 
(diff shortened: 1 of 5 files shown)


integration-mc1-16/build.gradle
```diff
@@ -100,6 +100,7 @@ dependencies {
     // The userdev artifact is a special name and will get all sorts of transformations applied to it.
     minecraft 'net.minecraftforge:forge:1.16.5-36.0.22'
     implementation 'thedarkcolour:kotlinforforge:1.9.0'
+    implementation project(':core')
```
jrddunbr, line 103: This is required for the Minecraft code to know about the Core project classes (ie, IProcess and the sim code)

jrddunbr:
It might be worth making a special API package where you don't need to actually include the core simulation code, so that other mods know how to interface with stuff properly. Thing is, that API would also need to present MNA primitives because we currently don't really have a great abstraction layer for when that isn't handled by the MNA we have; it's relatively tightly implemented and we may need another layer here.

jrddunbr:
Cameron and I talked about this and we think that we want to change how this abstraction happens. Going to rewrite.

PR #141, opened 2022-02-08
Various refactoring after most recent changes

jrddunbr:
This makes some various refactoring changes; adds a retry limiter to the `HttpPlot` methods so that it doesn't fill your logs all the time, and made better exception handlers for places in the code where I found `!!` (although I may not have gotten all of them).

I also removed all local instances of loggers and pointed them at the global Eln2 logger instance in the Eln2 mod class.

### DIFF
--- a/src/main/java/org/eln2/mc/common/blocks/CellTileEntity.kt
+++ b/src/main/java/org/eln2/mc/common/blocks/CellTileEntity.kt
@@ -10,6 +10,7 @@ import net.minecraft.world.level.Level
 import net.minecraft.world.level.block.entity.BlockEntity
 import net.minecraft.world.level.block.state.BlockState
 import org.apache.logging.log4j.LogManager
+import org.eln2.mc.Eln2
 import org.eln2.mc.common.*
 import org.eln2.mc.common.cell.*
 import java.util.*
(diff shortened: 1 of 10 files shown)


src/main/java/org/eln2/mc/common/blocks/CellTileEntity.kt
```diff
@@ -140,11 +140,12 @@ class CellTileEntity(var pos : BlockPos, var state: BlockState): BlockEntity(Blo
...
         }
 
-        logger.info("Remove nanoseconds: $nanoTime")
+        Eln2.LOGGER.info("Remove nanoseconds: $nanoTime")
 
         cell.graph.destroyAndRemove()
 
-        val t = thread (start = true) {
+        thread (start = true) {
+            // TODO: We're creating a thread... and then HttpPlot is creating it's own thread for every item in the list. This should be cleaned up, it won't scale once we release.
```
jrddunbr, line 148: We're creating a thread... and then HttpPlot is creating it's own thread for every item in the list. This should be cleaned up, it won't scale once we release.

I did make it so that it wouldn't make child threads if the HttpPlot client wasn't connected which saves creating useless threads. This should help performance if you are not using the debug harness.

src/main/java/org/eln2/mc/common/blocks/HttpPlot.kt
```diff
@@ -5,15 +5,20 @@ import net.minecraft.core.Direction
 import org.apache.http.client.methods.HttpPost
 import org.apache.http.entity.StringEntity
 import org.apache.http.impl.client.HttpClients
+import org.eln2.mc.Eln2.LOGGER
 import org.eln2.mc.common.cell.CellGraph
 import kotlin.concurrent.thread
 
 object HttpPlot {
+    var attempts = 0
+    const val ATTEMPT_COUNT = 3
```
jrddunbr, line 14: We can change this to 1 or 2 but 3 seems reasonable.

Caeleron (approved):

empireu (approved):

PR #144, opened 2022-02-12
Adds libelectric dependency

jrddunbr:
Yep, this works. I was able to run this both in the development environment as well as the actual Minecraft launcher.

Please note not to forget to grab the Kotlin For Forge dependency if you pull this in via the Minecraft launcher.


empireu (approved):
Fair enough.

PR #145, opened 2022-02-12
Adds server test

jrddunbr:
This creates a new server test to make sure that we're not breaking the server with our changes.

It only checks the server in development mode, so we may also want to make sure that standalone servers also work in the future (but this is mostly the same and a decent indicator).


jrddunbr:
GitHub is _really_ slow. I could probably do this way faster in a Lambda or something with actual processing power.

.github/workflows/run_server.sh
```diff
@@ -0,0 +1,49 @@
...
+echo "Printing server latest log:"
+cat ./run/logs/latest.log
+
+# GitHub is really slow so we're just going to check it generated chunks...
+echo "Checking for successful chunkloading (\"Preparing spawn area: \" messages)"
+SPAWN_AREA=$(grep "Preparing spawn area: " < ./run/logs/latest.log | wc -l)
+
+if [ "$SPAWN_AREA" != "0" ]; then
+    echo "Server started successfully! :)"
+    # Good enough.
```
Grissess, line 43: I can think of a number of crash-after-loads where this _isn't_ good enough, but looking for errors is fruitless because of how many of them are logged after being caught, so...

You know what, sure, good enough :P

Grissess (approved):
other than possible missing a few errors (discussed above) which can be looked into later, LGTM for now. Let's just get the infrastructure up, I s'pose :)

.github/workflows/run_server.sh
jrddunbr: So the reason I didn't try doing that is 1) how the heck am I going to run a Minecraft _client_ in a GL-less context where I barely have access to `sudo` and 2) GitHub is so slow that world generation can take probably upwards of 5-10 minutes and I don't want to wait that long when most errors will be found in the first 30 seconds of the classloader.

PR #147, opened 2022-02-13
Various Cleanup

jrddunbr:
* Cleans up logger instances
    * I keep finding these; @empireu, you do realize that a log4j logger is thread safe and designed for it, right? Using one static logger instance is going to be a lot better than building new ones all the time.
> Log4j components are designed to be used in heavily multithreaded systems.
https://logging.apache.org/log4j/1.2/faq.html#a1.7
* Fixed a bug in Circuit Explorer Open Packet implementation (See L42)
* Set some linter overrrides for situations where they didn't help anything.
* Converted Networking packet registration to use an enum instead of an autoincremented digit to prevent issues in the future if packets are added out-of-order (id's are now static to the type in the enum declaration)

### DIFF
--- a/src/main/java/org/eln2/mc/common/packets/clientToServer/CircuitExplorerOpenPacket.kt
+++ b/src/main/java/org/eln2/mc/common/packets/clientToServer/CircuitExplorerOpenPacket.kt
@@ -5,7 +5,6 @@ import net.minecraft.server.level.ServerPlayer
 import net.minecraft.world.phys.BlockHitResult
 import net.minecraft.world.phys.HitResult
 import net.minecraftforge.network.NetworkEvent
-import org.apache.logging.log4j.LogManager
 import org.eln2.mc.Eln2
 import org.eln2.mc.client.gui.CellInfo
 import org.eln2.mc.common.Networking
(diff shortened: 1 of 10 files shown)


src/main/java/org/eln2/mc/common/packets/clientToServer/CircuitExplorerOpenPacket.kt
```diff
@@ -30,15 +31,15 @@ class CircuitExplorerOpenPacket() {
...
+                Eln2.LOGGER.info("server did not hit")
                 return
             }
 
             val blockHit = pickResult as BlockHitResult
             val tile = player.level.getBlockEntity(blockHit.blockPos) as CellTileEntity?
 
             if(tile !is CellTileEntity){
-                Eln2.LOGGER.info("it is not the tile, it is $tile")
+                Eln2.LOGGER.info("it is not the tile, it is ${player.level.getBlockEntity(blockHit.blockPos)}")
```
jrddunbr, line 42: Calling out this bugfix, the type of `val tile` is either `CellTileEntity` or `null`, and if it isn't a cell tile entity, this will always print null. Now, it will print the pre-casted type.

empireu (approved):

PR #151, opened 2022-02-14
Feature/humanization

jrddunbr:
* Adds item textures
* Adds localizations


jrddunbr:
@Grissess can you un-fuck this and rebase it off dev? I forgot to pull from dev and it's angry now..

PR #154, opened 2022-02-14
Remove external library and add as a sourceDep instead

Grissess:


jrddunbr (approved):
LGTM as long as build passes.

jrddunbr:
```
Run ./.github/workflows/run_server.sh
tmux installed, setting eula and starting server
Starting Minecraft Server
Killing session
Printing server latest log:
cat: ./run/logs/latest.log: No such file or directory
Checking for successful chunkloading ("Preparing spawn area: " messages)
./.github/workflows/run_server.sh: line 39: ./run/logs/latest.log: No such file or directory
Server failed to load :(
Error: Process completed with exit code 1.
```
Hmm. Looks like it failed to run the server. Too bad I don't log stdout from tmux.. (is that possible?)

Grissess:
Give it a moment, this is nearly as bad a fight with Gradle as it was with ChatAge. It looks like the implementation of sourceControl in Gradle is by and large untested and poorly documented

jrddunbr (changes requested):
Build failure

jrddunbr:
We selected that the name of the package would be `libage`, so I think that if we have binaries that fixes this issue without the need for source code imports?

jrddunbr:
Possibly going to be replaced by #160 if merged, but worth noting that this process in general (Forge Gradle) is more flaky than it probably should be. Bloxgate identified with #160 that there's an issue with the classpath where it will load both a new and a old version of annotations (13 and 26) into the classpath instead of not using just the newer version like Gradle expects.

PR #158, opened 2022-02-14
Working Circuits

jrddunbr:
Thanks @empireu for this pull request!

This probably gets circuits into working order. I'd have to check.


src/main/java/org/eln2/mc/client/gui/PlotterScreen.kt
```diff
@@ -121,31 +133,37 @@ class PlotterScreen(private val cells : ArrayList<CellInfo>) : Screen(TextCompon
...
             }
         }
 
+        /*
+        if(drawList.isNotEmpty()){
+            drawNanoseconds = measureNanoTime {
+                pPoseStack.blitMultiple(drawList)
+            }
+        }*/
+
```
jrddunbr, line 154: What's this for?

src/main/java/org/eln2/mc/common/blocks/ConduitBlock.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    override fun getRenderShape(pState: BlockState): RenderShape {
+        return RenderShape.MODEL
+    }
+
+
+
+
+}
```
jrddunbr, line 61: What's this?

jrddunbr: Side note: [git stash](https://www.atlassian.com/git/tutorials/saving-changes/git-stash) might be useful to you.

src/main/java/org/eln2/mc/common/CommonEvents.kt
```diff
@@ -1,17 +1,30 @@
...
+    }
+
+    @SubscribeEvent
+    fun onChat(event : ServerChatEvent){
+        Eln2.LOGGER.info("Chat: ${event.message}")
+        if(event.message == "build"){
+            CellGraphManager.getFor(event.player.level as ServerLevel).graphs.values.forEach{ it.build() }
         }
     }
 }
```
jrddunbr, line 30: What is the purpose of this exactly? I ran the command and it doesn't seem to do anything too useful? Probably for debugging?

src/main/java/org/eln2/mc/common/cell/CellBase.kt
```diff
@@ -46,8 +50,33 @@ abstract class CellBase(val pos : BlockPos) {
...
+        return "N/A"
+    }
+
+    /**
+     * This method is called before the Circuit is being rebuilt.
+    */
+    abstract fun clearForRebuild()
+
+    /**
+     * This method is used to return the component and the pin for a remote cell to connect to.
```
jrddunbr, line 73: add `@param`s for the input and `@return` for the return type

src/main/java/org/eln2/mc/extensions/PoseStackExtensions.kt
```diff
@@ -0,0 +1,35 @@
...
+            builder.vertex(matrix, pX1, pY2, 0f).uv(0f, height).endVertex()
+            builder.vertex(matrix, pX2, pY2, 0f).uv(width, height).endVertex()
+            builder.vertex(matrix, pX2, pY1, 0f).uv(width, 0f).endVertex()
+            builder.vertex(matrix, pX1, pY1, 0f).uv(0f, 0f).endVertex()
+            builder.end()
+        }
+
+        BufferUploader.end(builder)
+    }
+}
```
jrddunbr, line 35: Could you comment about what this is going at the class/function level?

jrddunbr (commented):
just some comments

src/main/java/org/eln2/mc/extensions/PoseStackExtensions.kt
empireu: this was supposed to be an optimized way of drawing textures to the screen, but it unfortunately does not work and can be removed.

src/main/java/org/eln2/mc/client/gui/PlotterScreen.kt
empireu: this was used by the optimized drawing method. It can be removed.

src/main/java/org/eln2/mc/common/blocks/ConduitBlock.kt
empireu: this was a test of the voxel engine. We can nicely create cables using this. Unfortunately, texturing them will be a bit troublesome. Another method might be better,

src/main/java/org/eln2/mc/common/CommonEvents.kt
empireu: It was used before I rebuild the circuits on each edit. It can be removed.

jrddunbr:
I'm just going to merge this because I'll probably have to analyze the current situation as a whole and find a path forwards.

Thank you for your contributions.

PR #160, opened 2022-02-17
Add JitPack library support

jrddunbr:
So this builds just fine. But if you try to run the client, you get an error about Jetbrains annotations.

Weirdest thing ever. Otherwise, this would be how we move to JitPack. (See https://github.com/age-series/libage/issues/13 for context)


jrddunbr:
```
> Task :runClient FAILED
Exception in thread "main" java.lang.module.ResolutionException: Modules org.jetbrains.annotations and annotations export package org.intellij.lang.annotations to module terminalconsoleappender
        at java.base/java.lang.module.Resolver.resolveFail(Resolver.java:901)
        at java.base/java.lang.module.Resolver.failTwoSuppliers(Resolver.java:815)
        at java.base/java.lang.module.Resolver.checkExportSuppliers(Resolver.java:755)
        at java.base/java.lang.module.Resolver.finish(Resolver.java:380)
        at java.base/java.lang.module.Configuration.<init>(Configuration.java:140)
        at java.base/java.lang.module.Configuration.resolveAndBind(Configuration.java:494)
        at java.base/java.lang.module.Configuration.resolveAndBind(Configuration.java:298)
        at cpw.mods.bootstraplauncher@1.0.0/cpw.mods.bootstraplauncher.BootstrapLauncher.main(BootstrapLauncher.java:137)
```

jrddunbr:
I'm going to ignore the server test failure because I think the issue is that the GitHub runners are using an outdated gradle cache and that it needs to be flushed. I ran the same script locally on my box and it worked on the first try so unless there's some really weird Gradle issue because of my local cache, it's probably not actually an issue.

Issue #162, opened 2022-02-17
Cell Refactoring


jrddunbr:
Cameron and I have some thoughts about how we should refactor Cell a bit.

Cameron thinks that cells should directly work with the Cell Graph Manager in order to get added to cell groups.

Cell Graphs will likely be sim-centric; eg. we might have a `CircuitCellGraph`, `ThermalCellGraph`, `ShaftCellGraph`, etc.

jrddunbr:
Cameron also wants to investigate whether this part of Minecraft is threaded; @bloxgate do you have any idea how this might work?

jrddunbr:
Cell Provider refactoring will also probably be required so that a function is called on the cell and the cell returns the connection mapping with the type of the simulator connections on each side.

bloxgate:
> Cameron also wants to investigate whether this part of Minecraft is threaded; @bloxgate do you have any idea how this might work?

Last time I really looked into it there were only the game and network threads. That said, there's nothing really forbidding us from making another thread for our own stuff.

PR #163, opened 2022-02-17
Minor refactors [Ready for Review]

jrddunbr:
@empireu These are the minor changes that Cameron suggested (mostly around not using `if (condition) return` statements).

There's also #162 which captures some larger concerns that he will probably look into later but are not relevant to this PR.

We also don't really see the utility in the `@In` annotation and it's somewhat incorrect in some places where the code is called on both the client and the server side; we should simply replace it with javadoc documentation around where we want to have these things indicated to developers.


jrddunbr:
Codacy can be ignored here. It's angry that I moved a line around with a generic exception.

empireu (approved):

PR #164, opened 2022-02-18
Feature/plotter updates [Pending refactoring into new PR's]

jrddunbr:
Plotter updates


src/main/java/org/eln2/mc/utility/DataLabelBuilder.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    fun <T> enumerateOn(enumeration: List<T>, editor: (T, DataLabelBuilder) -> Unit) : DataLabelBuilder{
+        for(x in enumeration){
+            editor(x, this)
+        }
+
+        return this
+    }
+}
```
jrddunbr: I mentioned this before but for the record I don't feel like this is abstract enough and that the colors should be split away from the text generation for this. I want to be able to print out unit types and _then_ apply colors with a function like perhaps `applyColorToText(text: String, color: Color): MinecraftColoredText` or something.

src/main/java/org/eln2/mc/common/cell/types/WireCell.kt
```diff
@@ -1,27 +1,61 @@
...
-        Eln2.LOGGER.info("Wire $pos update -> connections: $connectionsChanged graphs: $graphChanged")
+    private val neighbourToResistorLookup = HashMap<AbstractCell, Resistor>()
+
+    //todo: bogus
+    override fun createDataPrint(): DataLabelBuilder {
+        val builder = DataLabelBuilder()
+
+        connections.forEachIndexed { index, remoteCell ->
+            builder.siEntry("Connection ${index}: ", "A", neighbourToResistorLookup[remoteCell]!!.current, McColors.red)
+        }
```
jrddunbr: Probably want to fix this.

src/main/java/org/eln2/mc/common/cell/types/GroundCell.kt
```diff
@@ -1,27 +1,55 @@
...
 
-    override fun update(connectionsChanged: Boolean, graphChanged: Boolean) {
-        Eln2.LOGGER.info("Ground $pos update -> connections: $connectionsChanged graphs: $graphChanged")
+    private val neighbourToResistorLookup = HashMap<AbstractCell, Resistor>()
+
+    // todo: bogus
+    override fun createDataPrint(): DataLabelBuilder {
+        val current = connections.sumOf { (componentForNeighbour(it).component as Resistor).current }
+        return DataLabelBuilder().amps(current, McColors.red)
     }
```
jrddunbr: Probably want to fix this.

src/main/java/org/eln2/mc/common/blocks/BlockRegistry.kt
```diff
@@ -55,4 +52,6 @@ object BlockRegistry {
     val WIRE_CELL = registerCellBlock("wire", eln2Tab) { WireCellBlock() }
     val VOLTAGE_SOURCE_CELL = registerCellBlock("voltage_source", eln2Tab) { VoltageSourceCellBlock() }
     val GROUND_CELL = registerCellBlock("ground", eln2Tab) { GroundCellBlock() }
+    val CAPACITOR_CELL = registerCellBlock("capacitor", eln2Tab) { CapacitorCellBlock() }
+    val INDUCTOR_CELL = registerCellBlock("inductor", eln2Tab) { InductorCellBlock() }
```
jrddunbr, line 56: This should be it's own pull request. Please don't add fixes, a naming refactor, plus plotter changes in one big request.

jrddunbr (commented):
I submitted this so I can't reject this (lol) but this needs to be split into at least 3 pull requests and rebased against `dev` so that it's not including already-commited code. This has a lot of changes compared to dev simply because of the work I've done since this was submitted but having old commits on it that you made that have been merged makes this harder to review.

jrddunbr:
I broke out the inductor and capacitor code into it's own pull request.

src/main/java/org/eln2/mc/common/cell/AbstractCell.kt
```diff
@@ -0,0 +1,87 @@
...
+    */
+    @In(Side.LogicalServer)
+    open fun update(connectionsChanged : Boolean, graphChanged : Boolean){}
+
+    /**
+     * This method returns the text that will be displayed in the Circuit Explorer.
+    */
+    open fun createDataPrint() : DataLabelBuilder{
+        return DataLabelBuilder()
+    }
```
jrddunbr: This used to be a `String` and I kinda liked it better like that, but this also works.

My only thing is that we might support `IDetail` from `libage` here in the future so being able to return a `String` would be cool too.

Since this is only used for the plotter right now I'm not too attached but I feel like being able to dump the state of the graphs in a plain-text format would be really useful (perhaps we should just write a toString implementation?).

src/main/java/org/eln2/mc/utility/DataLabelBuilder.kt
empireu: Has been addressed in the latest commit. More polishing can be carried out at a later date.

src/main/java/org/eln2/mc/extensions/DataBuilderExtensions.kt
```diff
@@ -0,0 +1,111 @@
...
+import org.eln2.libelectric.sim.electrical.mna.component.Capacitor
+import org.eln2.libelectric.sim.electrical.mna.component.Inductor
+import org.eln2.libelectric.sim.electrical.mna.component.Resistor
+import org.eln2.libelectric.sim.electrical.mna.component.VoltageSource
+import org.eln2.mc.utility.*
+import kotlin.math.abs
+
+object DataBuilderExtensions {
+    private class InternationalSystemEntry(override val label : String, val number : Double, val decimals : Int, val unit : String)
+        : DataBuilder.Entry(label, SuffixConverter.convert(abs(number), unit, decimals))
```
jrddunbr, line 12: Decimals are nice but Cameron and I were thinking we should use significant digits instead like old Electrical Age did.

src/main/java/org/eln2/mc/extensions/DataBuilderExtensions.kt
```diff
@@ -0,0 +1,111 @@
...
+        return entry(InternationalSystemEntry("Energy", energy, 2, "J"))
+    }
+
+    fun DataBuilder.farads(capacitance : Double) : DataBuilder {
+        return entry(InternationalSystemEntry("Capacitance", capacitance, 2, "F"))
+    }
+
+    fun DataBuilder.henry(inductance : Double) : DataBuilder {
+        return entry(InternationalSystemEntry("Inductance", inductance, 2, "H"))
+    }
```
jrddunbr, line 40: Why can't these be strings?

src/main/java/org/eln2/mc/extensions/DataBuilderExtensions.kt
```diff
@@ -0,0 +1,111 @@
...
+        return this
+    }
+
+    fun DataBuilder.withAmpsLabelColor() : DataBuilder {return this.withLabelColor(McColors.red)}
+    fun DataBuilder.withVoltsLabelColor() : DataBuilder {return this.withLabelColor(McColors.cyan)}
+    fun DataBuilder.withOhmsLabelColor() : DataBuilder {return this.withLabelColor(McColors.yellow)}
+    fun DataBuilder.withJoulesLabelColor() : DataBuilder {return this.withLabelColor(McColors.purple)}
+    fun DataBuilder.withHenryLabelColor() : DataBuilder {return this.withLabelColor(McColor(255u, 170u, 80u))}
+
+    fun DataBuilder.withElectricalValueColor() : DataBuilder {return this.withValueColor(McColors.green)}
```
jrddunbr, line 110: Why can't these be operations on `String`s that can return this data type?

jrddunbr:
And closed. We've pulled in everything on this front.

PR #175, opened 2022-02-19
Feature/wthit

jrddunbr:
Adds WTHIT support


bloxgate (approved):

jrddunbr:
Bloxgate and I peer programmed this whole thing so it should be high quality (thanks for the hlep!)

PR #182, opened 2022-02-19
Removes In, Adds util methods

jrddunbr:
* Removes the @In annotation
* Adds some utility methods
* Exposes the simulator solve state in Cell


jrddunbr:
Lol that poor developer that gets pinged for every `@In` call...

Caeleron (approved):

PR #183, opened 2022-02-19
Feature/plotterfixes

### LINKED ISSUE #184: Replace all generic Exception() with Kotlin error()
In multiple places in the code we have a block which basically goes `throw Exception("A message for the developer")` and it has been suggested that we move over to using the more Kotlin-friendly `error("A message for the developer") format.

* Replace all instances of `throw Exception(...)` with `error(...)`

jrddunbr:
This updates the plotter code and some associated networking stack code with it.


src/main/java/org/eln2/mc/utility/McColor.kt
```diff
@@ -0,0 +1,67 @@
+package org.eln2.mc.utility
+
+import net.minecraft.network.FriendlyByteBuf
+
+class McColor(val r : UByte, val g : UByte, val b : UByte, val a : UByte) {
+    constructor(r : Int, g : Int, b : Int, a : Int) : this(
+        if (r < 256 && r > -1) r.toUByte() else throw Exception("R is not within bounds."),
```
Caeleron: Grissess mentioned in one of my PRs that Codacy likes the pattern `error(<msg>)` rather than `throw Exception(<msg>)`. Perhaps we can swap these over too?

jrddunbr: I've added #184 to the issue tracker to replace all of these instances, since I'm sure there are plenty of them.

jrddunbr:
Fixes #184 as well.

jrddunbr:
I also deleted the SuffixConverter class because it is not going to be used anymore.

jrddunbr:
Cameron said he's good with this.

PR #185, opened 2022-02-19
Adds configuration and statistics

### LINKED ISSUE #172: Mod Analytics
- [ ] Opt-Out Config
- [ ] Initial Analytics packet
- [ ] Block place/remove counter
- [ ] Final Analytics packet

Initial packet:
* UUID (not the Minecraft one, just a unique one generated and placed in `.minecraft/age-series-uuid.txt`.
* Operating System Language
* Game Language

Final Packet:
* UUID (not the Minecraft one, just a unique one generated and placed in `.minecraft/age-series-uuid.txt`.
* Blocks placed/removed (to gather utilization data to track if they are in a large modpack or actually being played)
* Average tick speed

We need to add an analytics endpoint so that we can see how the mod is performing and get user preferences (for example, languages we should support).

I want to do a thing where we send an initial analytics packet when they start the mod and another every time they leave a world, and add a counter that counts block placement and removals so that we can see if the mod is just loaded or if they actually play with it. My reasoning is that mods are often collected into major packs but that doesn’t really indicate whether they actually played it or enjoyed it.

Also, for languages, I want to get the system language and the game language because some people don’t use the game language settings because it’s poorly supported in some mods.

jrddunbr:
Adds:
* Configuration file - automatically makes configuration file if not generated already
* Analytics - sends analytics to analytics endpoint

The analytics are not completely implemented, someone can feel welcome to implement the following on top of it:
* Proper JSON serialization to the API (since the Json serializer is apparently broken)
* Send sim statistics at the end of the world (when saving)
* Send the actual client language (since I tried to and it crashes the game somehow)
* I'll need to fix my CAA records with the domain name registrar so that I can use the proper DNS name.

The code that I used to collect analytics will collect IP, uuid (unique per config, not the Mojang one), system language, and will later collect one or two other things. See the existing disclaimer for Eln1 for now because it's the same stuff really.

~The actual source code will be uploaded to a different repository. I don't currently have CDK for the API and made it by hand in my AWS account for this stuff.~ Done.

### DIFF
--- a/src/main/java/org/eln2/mc/common/Configuration.kt
+++ b/src/main/java/org/eln2/mc/common/Configuration.kt
@@ -1,0 +1,54 @@
+package org.eln2.mc.common
+
+import com.charleskorn.kaml.Yaml
+import kotlinx.serialization.Serializable
+import org.eln2.mc.Eln2.LOGGER
+import java.io.File
+import java.util.*
+
+object Configuration {
+    private val CONFIG_FILE: File = File("config/electrical_age.yaml")
+    var config: ElectricalAgeConfiguration = ElectricalAgeConfiguration()
+
+    fun loadConfig(configFile: File = CONFIG_FILE) {
+        try {
+            if (configFile.isFile) {
+                LOGGER.info("[Electrical Age] Reading config from ${configFile.absoluteFile}")
+                config = Yaml.default.decodeFromStream(ElectricalAgeConfiguration.serializer(), configFile.inputStream())
+            } else {
+                config = ElectricalAgeConfiguration()
+                saveConfig()
+            }
+        } catch (e: Exception) {
+            LOGGER.error("Electrical Age had an issue with loading the configuration file, please check the file for errors.")
+            LOGGER.error("Check that 1) You have valid YAML 2) the config directives are spelled correctly (see documentation)")
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


src/main/java/org/eln2/mc/common/Configuration.kt
```diff
@@ -0,0 +1,54 @@
...
+ * ElectricalAgeConfiguration
+ *
+ * NOTE: ALL FIELDS _MUST_ have default values when called, since the user will likely want a sane default!
+ * They may be blank, but MUST be present. Thanks!
+ */
+@Serializable
+data class ElectricalAgeConfiguration (
+    var enableAnalytics: Boolean = true,
+    // TODO: Replace with stats.age-series.org (but needs CAA certificates)
+    var analyticsEndpoint: String = "https://ingz5drycg.execute-api.us-east-1.amazonaws.com/",
```
jrddunbr, line 52: This is because the CAA certificate record is broken and I can't get Amazon to recognize it.

Might need to port `age-series.org` out of the current registrar and into a different one.

bloxgate (approved):
LGTM, though I think a message in chat at first launch mentioning we do this might be a good idea.

As to the broken JSON parsing, have you taken a look at [cbeust/klaxon](https://github.com/cbeust/klaxon)? I'm not 100% sure it's built to use kotlinx-serialization, but I've had good experiences with it in some personal projects.

PR #186, opened 2022-02-20
Feature/models textures

jrddunbr:
Adds various models and textures:

* 12 Volt Battery
* Cable
* Knife Switch
* Light
* Solar Light
* 1x1 Solar Panel


src/main/java/org/eln2/mc/common/cell/types/12VBatteryCell.kt
```diff
@@ -0,0 +1,75 @@
...
+    /*
+    *   V -> local voltage source.
+    *   C -> remote components.
+    *
+    *   C V C
+    */
+
+    private lateinit var source : VoltageSource
+
+    override fun clearForRebuild() {
```
bloxgate, line 23: This is a problem in a few other places as well, but we should restore the potentials and other information from the save file when rebuilding.

As another example: the voltage source doesn't persist changes to voltage across world loads.

bloxgate (approved):

PR #196, opened 2022-02-23
Restructuring the mod

jrddunbr:
Hot Take - We restructure the mod like this.


src/main/java/org/eln2/mc/common/cell/ElectricalComponentConnection.kt
```diff
@@ -0,0 +1,6 @@
+package org.eln2.mc.common.cell
+
+import org.ageseries.libage.sim.electrical.mna.component.Component
+
+// TODO: This should be abstracted by the relevant simulator with a generic to allow for more than just a Electrical component.
+data class ElectricalComponentConnection(val component: Component, val index : Int)
```
jrddunbr, line 6: I renamed this to be more representative of what it is; in the future we should abstract this per MNA and then the mna will register a new subclass that it uses for transport of the feature.

This interface should probably come from libsim or libsim-minecraft to allow for MNA's to prepare for that sort of inbound structure?

src/main/java/org/eln2/mc/extensions/DirectionExtensions.kt
```diff
@@ -0,0 +1,9 @@
+package org.eln2.mc.extensions
+
+import net.minecraft.core.Direction
+
+// Some helper functions for rotations in Direction
+fun Direction.getFront(dir: Direction) = dir
+fun Direction.getBack(dir: Direction) = dir.opposite
+fun Direction.getLeft(dir: Direction) = dir.counterClockWise
+fun Direction.getRight(dir: Direction) = dir.clockWise
```
jrddunbr: lol meant to delete this. Was messing with some restructuring of direction.

jrddunbr:
This is still technically waiting for review at some point.

PR #197, opened 2022-02-23
Implement single-value adjustables

### LINKED ISSUE #171: Adjustable devices
The following needs a GUI to configure values for testing:

- [x] Voltage Source (volts)
- [x] Resistor (ohms)
- [x] Capacitor (microfarads)
- [x] Inductor (millihenries)

bloxgate:
Closes #171


src/main/resources/assets/eln2/lang/en_us.json
```diff
@@ -29,13 +29,13 @@
...
     "waila.eln2.voltage": "Voltage",
     "waila.eln2.current": "Current",
     "waila.eln2.resistance": "Resistance",
     "waila.eln2.capacitance": "Capacitance",
     "waila.eln2.inductance": "Inductance",
     "waila.eln2.energy": "Energy",
     "waila.eln2.power": "Power",
-    "waila.eln2.mode": "Mode"
+    "waila.eln2.mode": "Mode",
+    "gui.eln2.save": "Save"
```
jrddunbr: I'd add some whitespace between these, as I've been trying to space them out when they are different components to be translated (might make it easier on our translators)

src/main/java/org/eln2/mc/common/network/serverToClient/SingleDoubleElementGuiOpenPacket.kt
```diff
@@ -0,0 +1,49 @@
...
+            val ctx = supplier.get()
+
+            ctx.enqueueWork {
+                if (Dist.CLIENT == FMLEnvironment.dist) {
+                    handleClient(packet, Minecraft.getInstance())
+                }
+            }
+        }
+
+        @Suppress("UNCHECKED_CAST")
```
jrddunbr: Generally speaking I'm a fan of putting these directly adjacent to the statement if possible.

src/main/java/org/eln2/mc/Eln2.kt
```diff
@@ -36,4 +41,5 @@ object Eln2 {
             ModStatistics.sendAnalytics()
         }
     }
+
```
jrddunbr, line 49: ?

src/main/java/org/eln2/mc/client/screens/CapacitorCellScreen.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    private val textboxW = 192
+    private val textboxH = 16
+    private val textboxX = 32
+    private val textboxY = 48
+
+    private val buttonW = 60
+    private val buttonH = 18
+    private val buttonX = 96
+    private val buttonY = 128 - (buttonW / 2)
```
jrddunbr: Can these be const val? (or otherwise added to a (companion) object so we don't have dozens of instances of them?)

src/main/java/org/eln2/mc/client/screens/InductorCellScreen.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    private val textboxW = 192
+    private val textboxH = 16
+    private val textboxX = 32
+    private val textboxY = 48
+
+    private val buttonW = 60
+    private val buttonH = 18
+    private val buttonX = 96
+    private val buttonY = 128 - (buttonW / 2)
```
jrddunbr: ' '

src/main/java/org/eln2/mc/client/screens/ResistorCellScreen.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    private val textboxW = 192
+    private val textboxH = 16
+    private val textboxX = 32
+    private val textboxY = 48
+
+    private val buttonW = 60
+    private val buttonH = 18
+    private val buttonX = 96
+    private val buttonY = 128 - (buttonW / 2)
```
jrddunbr: ' '

src/main/java/org/eln2/mc/client/screens/VoltageSourceCellScreen.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    private val textboxW = 192
+    private val textboxH = 16
+    private val textboxX = 32
+    private val textboxY = 48
+
+    private val buttonW = 60
+    private val buttonH = 18
+    private val buttonX = 96
+    private val buttonY = 128 - (buttonW / 2)
```
jrddunbr: ' '

src/main/java/org/eln2/mc/common/blocks/cell/CapacitorCellBlock.kt
```diff
@@ -1,11 +1,60 @@
...
+                        return CapacitorCellContainer(pContainerId, pInventory, pPlayer, te)
+                    }
+                }
+
+                NetworkHooks.openGui(pPlayer as ServerPlayer, containerProvider, te.pos)
+                return InteractionResult.SUCCESS
+            }
+        }
+
+        return InteractionResult.SUCCESS
```
jrddunbr, line 58: Can we abstract this away from the implementation of each cell and provide a function with suppliers in the superclass?

src/main/java/org/eln2/mc/common/blocks/cell/InductorCellBlock.kt
```diff
@@ -1,11 +1,60 @@
...
+                        pPlayer: Player
+                    ): AbstractContainerMenu {
+                        return InductorCellContainer(pContainerId, pInventory, pPlayer, te)
+                    }
+                }
+
+                NetworkHooks.openGui(pPlayer as ServerPlayer, containerProvider, te.pos)
+                return InteractionResult.SUCCESS
+            }
+        }
```
jrddunbr, line 56: ' '

src/main/java/org/eln2/mc/common/blocks/cell/ResistorCellBlock.kt
```diff
@@ -1,11 +1,60 @@
...
+                        return ResistorCellContainer(pContainerId, pInventory, pPlayer, te)
+                    }
+                }
+
+                NetworkHooks.openGui(pPlayer as ServerPlayer, containerProvider, te.pos)
+                return InteractionResult.SUCCESS
+            }
+        }
+
+        return InteractionResult.SUCCESS
```
jrddunbr, line 58: ' '

src/main/java/org/eln2/mc/common/blocks/cell/VoltageSourceCellBlock.kt
```diff
@@ -1,11 +1,60 @@
...
+                        return VoltageSourceCellContainer(pContainerId, pInventory, pPlayer, te)
+                    }
+                }
+
+                NetworkHooks.openGui(pPlayer as ServerPlayer, containerProvider, te.pos)
+                return InteractionResult.SUCCESS
+            }
+        }
+
+        return InteractionResult.SUCCESS
```
jrddunbr, line 58: ' '

src/main/java/org/eln2/mc/common/cell/ISingleElementGuiCell.kt
```diff
@@ -0,0 +1,8 @@
+package org.eln2.mc.common.cell
+
+interface ISingleElementGuiCell<N : Number> {
+
+    fun getGuiValue(): N
+    fun setGuiValue(value: N)
+
+}
```
jrddunbr: Please add Javadoc comments to this interface

src/main/java/org/eln2/mc/common/cell/types/VoltageSourceCell.kt
```diff
@@ -23,7 +22,7 @@ class VoltageSourceCell(pos : BlockPos) : CellBase(pos) {
     *       C
     */
 
-    private lateinit var source : VoltageSource
+    lateinit var source: VoltageSource private set
```
jrddunbr: This feels dangerous to have a lateinit that is not private. Thoughts?

src/main/java/org/eln2/mc/common/containers/ContainerRegistry.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    @SubscribeEvent
+    fun clientSetup(event: FMLClientSetupEvent) {
+        event.enqueueWork {
+            MenuScreens.register(VOLTAGE_SOURCE_CELL_CONTAINER.get(), ::VoltageSourceCellScreen)
+            MenuScreens.register(RESISTOR_CELL_CONTAINER.get(), ::ResistorCellScreen)
+            MenuScreens.register(CAPACITOR_CELL_CONTAINER.get(), ::CapacitorCellScreen)
+            MenuScreens.register(INDUCTOR_CELL_CONTAINER.get(), ::InductorCellScreen)
+        }
+    }
```
jrddunbr: Suggest adding comments and perhaps an abstraction using suppliers to reduce boilerplate.

src/main/java/org/eln2/mc/common/containers/VoltageSourceCellContainer.kt
```diff
@@ -0,0 +1,32 @@
...
+import net.minecraft.server.level.ServerPlayer
+import net.minecraft.world.entity.player.Inventory
+import net.minecraft.world.entity.player.Player
+import org.eln2.mc.common.blocks.CellTileEntity
+import org.eln2.mc.common.cell.types.VoltageSourceCell
+import org.eln2.mc.common.containers.ContainerRegistry.VOLTAGE_SOURCE_CELL_CONTAINER
+import org.eln2.mc.common.network.Networking
+import org.eln2.mc.common.network.serverToClient.SingleDoubleElementGuiOpenPacket
+
+@Suppress("UNCHECKED_CAST")
```
jrddunbr: I'd prefer we do this at the statement level

src/main/java/org/eln2/mc/common/network/Networking.kt
```diff
@@ -24,19 +28,47 @@ object Networking {
...
+            ::CEOPacket,
+            CEOPacket::handle
+        )
+        channel.registerMessage(
+            PacketType.CIRCUIT_EXPLORER_CONTEXT_PACKET.id,
+            CECPacket::class.java,
+            CECPacket::encode,
+            ::CECPacket,
+            CECPacket::handle
+        )
```
jrddunbr: Since we've got the other ones at regular length and all indented, might as well remove the typealias at the top of the file.

jrddunbr (changes requested):
Decent.

src/main/java/org/eln2/mc/Eln2.kt
bloxgate: IntelliJ added this when it formatted the code on commit

src/main/java/org/eln2/mc/common/containers/ContainerRegistry.kt
```diff
@@ -0,0 +1,87 @@
...
+    )
+    private val CLIENTSIDE_GUI_QUEUE = mutableListOf<ContainerRegistryItem<*>>()
+
+    fun setup(bus: IEventBus) = CONTAINER_REGISTRY.register(bus)
+
+    data class ContainerRegistryItem<T: AbstractContainerMenu>(
+        val name: String,
+        val screenConstructor: (T, Inventory, Component) -> AbstractContainerScreen<T>,
+        val container: RegistryObject<MenuType<T>>
+    )
```
jrddunbr: I've been moving these out of the parent class because it helps discoverability of files in the project file tree and makes individual files easier to read (since it's somewhat shorter). They are also kinda an API-like thing so I'd like to comment them, but nothing we need to do before a refactor.

jrddunbr (approved):
This is good. I think we can get the duplication down once we've got a better idea of what our GUI structures are going to look like, but it's sufficient for now.

jrddunbr (changes requested):
Ahh darn look at that build failure. Client side classes. :/

src/main/java/org/eln2/mc/common/containers/ContainerRegistry.kt
bloxgate: This has been removed. The new register method only really exists to avoid having to explicitly type each registry object

src/main/java/org/eln2/mc/common/containers/ContainerRegistry.kt
bloxgate: I'm not 100% sure on how to abstract away the client registrations to be automated. In the meantime however, I've added some docs explaining the situation, and move the client registration to its own class.

src/main/java/org/eln2/mc/Eln2.kt
```diff
@@ -28,7 +31,9 @@ object Eln2 {
         ItemRegistry.setup(MOD_BUS)
         ContainerRegistry.setup(MOD_BUS)
 
-        MOD_BUS.register(ContainerRegistry) //client-side setup
+        if (Dist.CLIENT == FMLEnvironment.dist) {
```
jrddunbr, line 34: Does this work standalone?

jrddunbr (approved):
LGTM + good build.

src/main/java/org/eln2/mc/Eln2.kt
bloxgate: Yes, it should provide the physical side, determined at startup.

If we have issues, moving to a DistExecutor is simple enough - and would probably be good to do in the future anyway.

PR #199, opened 2022-02-26
Warn about analytics on first join

### LINKED ISSUE #187: Warn users that we collect analytics on first start
When the configuration file is generated for the first time _do not_ send an analytics packet and warn the user with a chat message that if they do not set it to false in the config that we will log basic information about their game's performance starting on the next run of the mod.

---

Currently we log 2 things that could be considered user-identifiers:
* uuid - a randomly generated UUID that is stored in their configuration for analytics
* IP address - the IP address that the server receives traffic from

These identifiers allow us to remove your data in the event of a GDPR (or if you just want to) request. Simply shoot a developer a message with your IP address(es) and/or UUID(s) from your config and we will delete _everything_ we have on your user.

The uuid field is also used as a dedup flag, and the IP address may be used for spam reduction purposes.

We also log the following currently (but may include more in the future - the source code is an authoritative reference):
* mod_id - what mod you're playing from Age Series (probably should also add version number?)
* game_lang - the language configured in the game
* os_lang - the language configured on the computer you are using (to find if we should be supporting more languages than people set in their Minecraft configuration)
* avg_sim_performance - average tick time of the MNA I guess? This is mostly a test field at this point but could combine all sims performance? idk.

Data is processed and stored in the AWS us-east-1 region, and besides the data collected itself, the various logs are configured to not be stored longer than a day on the components of the serverless architecture that we are using.

bloxgate:
Addresses #187


src/main/resources/assets/eln2/lang/en_us.json
```diff
@@ -39,5 +39,7 @@
     "waila.eln2.power": "Power",
     "waila.eln2.mode": "Mode",
 
-    "gui.eln2.save": "Save"
+    "gui.eln2.save": "Save",
+
+    "misc.eln2.acknowledge_analytics": "By default, Electrical Age 2 reports analytics to the mod author(s). If you wish to opt-out, please edit your config file. This message will not show again for 1 month."
```
jrddunbr: We should probably make this mod-agnostic in the future and refer to is as `Age Series Mods`.. granted we probably would have name conflicts. Ideally... this would be in `libage-minecraft` once it exists.

src/main/java/org/eln2/mc/common/CommonEvents.kt
```diff
@@ -38,4 +52,28 @@ object CommonEvents {
             }
         }
     }
+
+    @SubscribeEvent
+    fun onEntityJoinedWorld(event: EntityJoinWorldEvent) {
+        //Warn new users that we collect analytics
+        if (event.world.isClientSide && event.entity is Player) {
```
jrddunbr, line 57: So we're going to likely have both server and client side analytics so we're likely going to want to have _two_ configs eventually.

jrddunbr (commented):
I'm going to think about this. I think pestering after the second opt-out is probably undesired.

jrddunbr:
@bloxgate can we move this into the organization repository? It can't run the build tests from your personal GitHub. Just change remotes and push a branch.

bloxgate:
> @bloxgate can we move this into the organization repository? It can't run the build tests from your personal GitHub. Just change remotes and push a branch.

Done, though that feels like something we should fix for the future. We can't expect everyone opening a PR to be able to push to here.

src/main/java/org/eln2/mc/common/CommonEvents.kt
```diff
@@ -38,4 +52,28 @@ object CommonEvents {
...
+                val uuid: String = event.entity.uuid.toString()
+                val acknowledgements: AnalyticsAcknowledgementsData = Yaml.default.decodeFromStream(AnalyticsAcknowledgementsData.serializer(), acknowledgementFile.inputStream())
+                if (acknowledgements.entries.none { (k, _) -> k == uuid } || acknowledgements.entries[uuid]!! < (System.currentTimeMillis() / 1000L) - THIRTY_DAYS) {
+                    event.entity.sendMessage(TranslatableComponent("misc.eln2.acknowledge_analytics").withStyle(ChatFormatting.RED), Util.NIL_UUID)
+                    acknowledgements.entries[uuid] = System.currentTimeMillis() / 1000L
+                }
+
+                acknowledgementFile.writeText(Yaml.default.encodeToString(AnalyticsAcknowledgementsData.serializer(), acknowledgements))
+            }
+        }
```
jrddunbr: Need some checking for if the file is not writable or if the file read back is corrupted somehow (binary data, user messing with it, etc)

jrddunbr (changes requested):
This is decent, but I am worried about I/O issues.

src/main/java/org/eln2/mc/common/CommonEvents.kt
```diff
@@ -38,4 +50,53 @@ object CommonEvents {
             }
         }
     }
+
+    @SubscribeEvent
+    fun onEntityJoinedWorld(event: EntityJoinWorldEvent) {
+        //Warn new users that we collect analytics
+        if (event.world.isClientSide && event.entity is Player) {
+            if (Eln2.config.enableAnalytics)  {
+                val acknowledgementFile = FMLPaths.CONFIGDIR.get().resolve("ElectricalAge/acknowledgements.yml").toFile()
```
jrddunbr, line 59: Hmm. We're already putting things in `./config/electrical-age.yml` I think; should we have it as a similarly named file? Also, neat function.

src/main/java/org/eln2/mc/common/CommonEvents.kt
```diff
@@ -38,4 +50,53 @@ object CommonEvents {
...
+                if (!acknowledgementFile.exists()) {
+                    acknowledgementFile.parentFile.mkdirs()
+                    acknowledgementFile.createNewFile()
```
jrddunbr: These can both throw exceptions...

src/main/java/org/eln2/mc/common/CommonEvents.kt
```diff
@@ -1,16 +1,28 @@
...
 import net.minecraftforge.server.ServerLifecycleHooks
 import org.eln2.mc.Eln2
 import org.eln2.mc.common.cell.CellGraphManager
+import org.eln2.mc.utility.AnalyticsAcknowledgementsData
+import java.io.IOException
 
 @Mod.EventBusSubscriber
 object CommonEvents {
+
+    const val THIRTY_DAYS: Long = 2_592_000L
```
jrddunbr: Thirty days in... seconds?

src/main/java/org/eln2/mc/common/CommonEvents.kt
```diff
@@ -38,4 +50,53 @@ object CommonEvents {
...
+                        }
+                    }
+                }
+                val uuid: String = event.entity.uuid.toString()
+                val acknowledgements: AnalyticsAcknowledgementsData = try {
+                    Yaml.default.decodeFromStream(AnalyticsAcknowledgementsData.serializer(), acknowledgementFile.inputStream())
+                } catch (ex: IOException) {
+                    AnalyticsAcknowledgementsData(mutableMapOf())
+                }
+                if (acknowledgements.entries.none { (k, _) -> k == uuid } || acknowledgements.entries[uuid]!! < (System.currentTimeMillis() / 1000L) - THIRTY_DAYS) {
```
jrddunbr: I think 1000L is a magic number to convert milliseconds to seconds?

src/main/java/org/eln2/mc/utility/AnalyticsAcknowledgementsData.kt
```diff
@@ -0,0 +1,8 @@
+package org.eln2.mc.utility
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class AnalyticsAcknowledgementsData (
+    val entries: MutableMap<String, Long>
+)
```
jrddunbr: Missing newline

jrddunbr (changes requested):
Just a few nits here and there.

src/main/java/org/eln2/mc/common/CommonEvents.kt
bloxgate: Personally I think that any config files other than the main config (and sometimes even that) should go into a folder named for the mod. Perhaps a change from `ElectricalAge` to `electrical-age` would be warranted though, assuming that's the actual mod ID?

jrddunbr: lol, the modid is actually `eln2`. I don't know that modid's can actually have a dash in them due to Forge/Minecraft's ridiculous naming requirements. I've been calling it `electrical-age` and other names since `eln` isn't really a meaningful shortening for electrical age.

At some point we should probably review all of the names and make some decisions.

src/main/java/org/eln2/mc/common/CommonEvents.kt
```diff
@@ -38,4 +50,53 @@ object CommonEvents {
...
+                    }
+                }
+                val uuid: String = event.entity.uuid.toString()
+                val acknowledgements: AnalyticsAcknowledgementsData = try {
+                    Yaml.default.decodeFromStream(AnalyticsAcknowledgementsData.serializer(), acknowledgementFile.inputStream())
+                } catch (ex: IOException) {
+                    AnalyticsAcknowledgementsData(mutableMapOf())
+                }
+                if (acknowledgements.entries.none { (k, _) -> k == uuid } || acknowledgements.entries[uuid]!! < (System.currentTimeMillis()) - THIRTY_DAYS_AS_MILLISECONDS) {
+                    event.entity.sendMessage(TranslatableComponent("misc.eln2.acknowledge_analytics").withStyle(ChatFormatting.RED), Util.NIL_UUID)
```
jrddunbr, line 82: Does it need to be red? I think just standard text color is sufficient.. what do other mods do for startup messages? I feel like red should be reserved for things like crashes or critical errors.

jrddunbr (approved):
This looks good. Just two nitpick comments, but they don't need to be addressed now.

PR #203, opened 2022-03-05
Mark GUI packets as handled after work enqueued

### LINKED ISSUE #200: Unrecognized Packet Type eln2:main
Something's not entirely correct with how we're registering our simple network packets. I'm seeing an occasional mention of an unknown packet type for `eln2:main` when opening the adjustables' GUIs or the circuit explorer.

Likely we're just missing a registration step somewhere.

Edit to add: @empireu has informed me that we just need to make sure and set the packets as handled in their handlers and it should resolve itself.

bloxgate:
Fixes #200


jrddunbr (approved):
LGTM

PR #210, opened 2022-06-10
Add bare minimum Modrinth configuration

bloxgate:
Adds a gradle task that will publish if you have your `MODRINTH_TOKEN` set. Still manual at the moment

At the moment running this would create a version identical to the current 1.0.1 version, including the changelog.

ToDo:
- [ ] Automate this to publish on some criteria

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ buildscript {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "org.jetbrains.kotlin:kotlin-serialization:$kotlin_version"
         classpath "gradle.plugin.com.github.johnrengelman:shadow:7.1.1"
+        classpath group: 'com.modrinth.minotaur', name: 'Minotaur', version: '2.+'
     }
 }
 apply plugin: 'net.minecraftforge.gradle'
@@ -22,6 +23,7 @@ apply plugin: 'kotlinx-serialization'
 apply plugin: "com.github.johnrengelman.shadow"
 apply from: 'https://raw.githubusercontent.com/thedarkcolour/KotlinForForge/site/thedarkcolour/kotlinforforge/gradle/kff-3.3.2.gradle'
 apply plugin: 'eclipse'
+apply plugin: 'com.modrinth.minotaur'
 apply plugin: 'maven-publish'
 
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -195,3 +197,16 @@ compileTestKotlin {
...
+    token = System.getenv("MODRINTH_TOKEN")
+    projectId = "iTyN4Kv7"
+    versionNumber = "$project.version"
+    versionName = "eln2-$project.version"
+    versionType = "alpha"
+    uploadFile = shadowJar
+    gameVersions = [ "$mcVersion" ]
+    changelog = file('changes.md').text
+    dependencies {
+        required.project("ordsPcFz")
```
jrddunbr, line 210: Can we leave a comment here that it depends on KFF? Then I don't need to look up the ID.

gradle.properties
```diff
@@ -1,4 +1,6 @@
 # Sets default memory used for gradle commands. Can be overridden by user or command line properties.
 # This is required to provide enough memory for the Minecraft decompilation process.
 org.gradle.jvmargs=-Xmx3G
-org.gradle.daemon=false
\ No newline at end of file
+org.gradle.daemon=false
+
+mcVersion=1.18.2
```
jrddunbr, line 6: I see we are messing with the state of the universe in strange and dangerous ways /s :D

I recently did a commit that made the mod version number a constant in two places because the regex didn't work.. can you fix that too and perhaps add that here as well? I think it would make sense in the scope of this PR.

build.gradle
```diff
@@ -195,3 +197,16 @@ compileTestKotlin {
 }
 // However if you are in a multi-project build, dev time needs unobfed jar files, so you can delay the obfuscation until publishing by doing
 // publish.dependsOn('reobfJar')
+modrinth {
+    token = System.getenv("MODRINTH_TOKEN")
+    projectId = "iTyN4Kv7"
+    versionNumber = "$project.version"
```
jrddunbr, line 203: I left a comment further down about how this is not really globalized.

jrddunbr (approved):
Various thoughts but mostly there.

jrddunbr:
Recent merges mean that this PR needs to be rebased/refactored a bit.

jrddunbr (changes requested):
Needs to be rebased to actually be mergable at this point.

PR #227, opened 2022-08-10
Diode orientation

bloxgate:
I'm 100% sure there's a much better way to do this. However I'm not really familiar with the simulator enough to figure something out there.

- TwoPinCells no longer return true for any direction in their predicate
- The negative pin of a diode is now always the "front" of the diode

I've had some bizarre issues with saving while working on this, sometimes connecting circuits that were saved in a previous run of the world caused some stuff in the TEs to fail to save with a bunch of NPEs. I'm not sure it's related, but we may want to go clean up all our `!!` uses in cell a bit.

### DIFF
--- a/src/main/java/org/eln2/mc/common/blocks/CellTileEntity.kt
+++ b/src/main/java/org/eln2/mc/common/blocks/CellTileEntity.kt
@@ -1,6 +1,5 @@
 package org.eln2.mc.common.blocks
 
-import mcp.mobius.waila.api.IServerDataProvider
 import net.minecraft.core.BlockPos
 import net.minecraft.core.Direction
 import net.minecraft.nbt.CompoundTag
@@ -56,7 +55,7 @@ class CellTileEntity(var pos : BlockPos, var state: BlockState): BlockEntity(Blo
             return
         }
 
-        println("Cell is created at $position")
+        Eln2.LOGGER.debug("Cell is created at $position")
         cell = cellProvider.create(position)
         cell!!.tile = this
         cell!!.id = cellProvider.registryName!!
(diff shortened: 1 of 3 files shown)


jrddunbr:
Yeah, I'd advise for a refactor of Cell at some point to add cable type masks and better orientation code. It also doesn't really support the 3D six-node type stuff that Eln1 does.. although perhaps these are stretch goals of mine that I had - it would probably take a while or require more people on the devteam to figure it out in a reasonable time.

bloxgate:
Six nodes like in ELN1 are definitely a bit of a stretch goal, but I would
like to get the cables working 3D once we get the new models in.

Hopefully we can attract some more dev effort once we have something more
usable published.

jrddunbr:
Yeah, it's certainly a developer chicken-and-the-egg problem. I waited for 4 years and decided that was long enough. :/

Once you figure this out though, batteries, solar panels, and other "orientables" will start working more predictably and that's when the game starts to be interesting. I'd add lighting, (light) switches, and an electric furnace after that since those are the most immediately useful features for players to get captivated in the mod and showcase how it could be interesting to develop. But, sometimes it also comes down to what is interesting to you, and I'm not the one writing the code these days and so I'll just watch with anticipation :)

qrmcat (approved):

jrddunbr:
Any updates on this? Just curious.

bloxgate:
I've been quite busy with some IRL responsibilities for the past couple months. 

I believe that the diodes still have an issue with being randomly flipped when loading in with this change. That would require a change to how we save/load connection information though (and I seem to recall it affects a lot of other components too).

jrddunbr:
I'm going to guess that this problem either
1. Doesn't exist anymore
2. Is editing files that don't exist anymore

The change needs to be rebased on the most recent changes, regardless.

PR #238, opened 2026-03-14
Latest changes over the years

empireu:


jrddunbr:
I don't really care about Codacy passing for this but we probably should iron out the conflicts - we might have to do that merge locally?

Grissess (approved):
I'm agreed; Codacy is around 85% unhappy with constructor and parameter arities. The fact of the matter is that there will be data-heavy algorithms that will depend on constructing just such large aggregates of data.

As for the rest, this is pretty much a skim; the `dev` branch has always been set up for this kind of sandbox testing. Getting it to the main branch should be done with more care.

Grissess:
"remove failing tests"

PR #239, opened 2026-03-21
Tiny Wind Turbine Fix and GUI Icon and TItle Fixes

ParisGR123:
I replaced the missing textures of some items with their respective image and added their names in the en_us.json, i also made a minor fix to the wind turbine where placing it close to objects would cause the game to crash and the world to be lost along with it leaving air blocks around it (fixed by destroying delegates)


ParisGR123:
I also made an inventory model for the Standard Iron Straight Joint, thats all

